PR #5, opened 2017-01-17
fix timeout and package name

julianladisch:
On my laptop build fails because embedded postgres takes more than 2 minutes to load.

### DIFF
--- a/src/test/java/org/folio/moduserstest/RestVerticleTest.java
+++ b/src/test/java/org/folio/moduserstest/RestVerticleTest.java
@@ -1,3 +1,5 @@
+package org.folio.moduserstest;
+
 import io.vertx.core.DeploymentOptions;
 import io.vertx.core.Future;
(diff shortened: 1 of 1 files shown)


kurtnordstrom:
I apologize for not seeing this sooner. This means, of course, that any time we build, there's going to be a 180 second wait time for postgres to load, regardless of how fast the machine is?

julianladisch:
No. The faster the machine is the faster the test finishes. The timeout rule of junit fails the test when its executions time exceeds the timeout value. The rule only waits if the unit test hasn't finished yet. It does not wait if the unit test has finished.

kurtnordstrom:
Ah, I see. I have actually made a change to the test to load the pg client before the verticle gets run. Can you try it and let me know whether or not the issue still remains for you?

julianladisch:
The issue still remains. The total time for downloading and extracting postgres may take long when it is done for the first time.

PR #8, opened 2017-01-26
Removed obsolete /tenant endpoint from ModuleDescriptor.json

wafschneider:


wafschneider:
Also added tenantInterface key to direct Okapi to the correct tenant endpoint

PR #14, opened 2017-03-23
Make aspectj work in eclipse

julianladisch:
Use latest aspectj version that contains this fix
https://github.com/mojohaus/aspectj-maven-plugin/issues/18
to make it work in eclipse.


shale99:
checked locally

PR #16, opened 2017-03-27
DMOD-200 avoid multiple starts of embedded postgres

julianladisch:
use maven integration test phase and PostgresRunner


julianladisch:
This requires "DMOD-200 dropCreateDatabase(String database)": https://github.com/folio-org/raml-module-builder/pull/63/files

PR #19, opened 2017-04-04
FOLIO-532 fix dependency order causing jar to break

julianladisch:


dcrossleyau:
I did some verification of this today.

Built okay. Then followed Mike's quick deployment method in [ui-okapi-console/automation](https://github.com/folio-org/ui-okapi-console/tree/master/automation) and can get a list of users. Okay.

Then followed Kurt's full [mod-auth/TESTING.md](https://github.com/folio-org/mod-auth/blob/master/TESTING.md). Apart from the unrelated [DMOD-233](https://issues.folio.org/browse/DMOD-233), everything seemed to go well.

dcrossleyau:
While reviewing this PR's changes i did notice that a recent commit was accidently removed from the pom.xml file. I restored that. I tried to determine if anything else was missing. All seems to be fine.

MikeTaylor:
I should say that I've not used that "Okapi Console automation" approach
for some time now, since I switched to using Wayne's back-end black boxes.
So it's great if it's still useful to you, but be aware it's not maintained.

-- Mike.

PR #26, opened 2017-06-07
fix group examples: add id, fix usergroups key, add 2nd group

julianladisch:


dcrossleyau:
Merged. Thanks. The user-related RAMLs have moved to "raml-util" [folio-org/raml](https://github.com/folio-org/raml) so i have replicated there.

PR #27, opened 2017-06-07
user samples: use id for patronGroup

julianladisch:


dcrossleyau:
Merged. Thanks. The user-related RAMLs have moved to "raml-util" [folio-org/raml](https://github.com/folio-org/raml) so i have replicated there.

PR #30, opened 2017-06-08
remove ramls directory; has been moved into raml-util (submodule)

julianladisch:


dcrossleyau:
Waiting until the move to raml-util is settled.

There is similar in mod-login and log-permissions.

PR #62, opened 2018-04-19
Modtag 4

heikkilevanto:
Use a new raml-utils with tags in userdata. Add test cases for tags in users.

### DIFF
--- a/src/test/java/org/folio/moduserstest/RestVerticleIT.java
+++ b/src/test/java/org/folio/moduserstest/RestVerticleIT.java
@@ -137,13 +137,23 @@ public class RestVerticleIT {
     return future;
   }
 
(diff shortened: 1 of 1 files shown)


heikkilevanto:
Oops, looks like the pull request fails `Unrecognised URI, can't resolve this: file:/tmp/tags.schema: /tmp/tags.schema`. I think I know what that is all about.

heikkilevanto:
For some reason RMB copies the schemas into /tmp. On my machine it had done that during some previous experiment. So RMB could find it there all right. But I had forgotten to (re)add the inclusion of tags.schema into users.raml, so on any other machine it would fail. Also, there was a typo in a test (taglist instead of tagList), which caused another test failure.

PR #80, opened 2018-09-17
 MODUSERS-88: Extract exception processing into handle(...), fix CQL exception handling

julianladisch:


src/main/java/org/folio/rest/impl/UsersAPI.java
```diff
@@ -107,95 +109,75 @@ private CQLWrapper getCQL(String query, int limit, int offset) throws FieldExcep
...
-          Handler<AsyncResult<Response>> asyncResultHandler,
-          Context vertxContext) {
+      UsersGetOrder order, int offset, int limit, List<String> facets,
+      String lang, Map <String, String> okapiHeaders,
+      Handler<AsyncResult<Response>> asyncResultHandler,
+      Context vertxContext) {
+
     logger.debug("Getting users");
     try {
       CQLWrapper cql = getCQL(query,limit,offset);
```
julianladisch: getCQL may throw org.z3950.zing.cql.cql2pgjson.FieldException

PR #85, opened 2018-09-27
fix username's value may be verified as a numeric type

fluagen:
username's value may be verified as a numeric type that is cause MODUSERS-90 error.


adamdickmeiss:
Code incorrectly refers to userdata.json. It is now part of mod-users .. NOT the shared RAML. It is just bad practice that an OLD copy is also there.

fluagen:
OK, I will pay attention.

PR #93, opened 2018-11-06
MODUSERS-100: Get groups by invalid UUID hangs

julianladisch:


julianladisch:
The fix also disables calculating totalRecords because the totalRecords value is always thrown away and not needed by the /groups/{groupId} endpoint.

PR #119, opened 2019-06-12
Refactor to use PgUtil MODUSERS-129

adamdickmeiss:


adamdickmeiss:
Note that in two cases I was not able to use PgUtil where it should be possible because it returns 500 errors rather than 400-series.

src/main/java/org/folio/rest/impl/ProxiesForAPI.java
```diff
@@ -40,320 +36,116 @@
...
+        } else if (existsRes.result() == true) {
+          Errors existsError = ValidationHelper.createValidationErrorMessage(
+            "proxyFor", entity.getId(), "Proxy relationship already exists");
+          asyncResultHandler.handle(Future.succeededFuture(
+            PostProxiesforResponse.respond422WithApplicationJson(existsError)));
+        } else {
+          String id = entity.getId();
+          if (id == null) {
+            id = UUID.randomUUID().toString();
+            entity.setId(id);
```
julianladisch: Drop this else-clause. This is done by PgUtil.post: https://github.com/folio-org/raml-module-builder/blob/v25.0.0/domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java#L505

src/main/java/org/folio/rest/impl/ProxiesForAPI.java
```diff
@@ -40,320 +36,116 @@
...
-        String message = logAndSaveError(e);
-        asyncResultHandler.handle(Future.succeededFuture(
-                    PostProxiesforResponse.respond500WithTextPlain(
-                            getErrorResponse(message))));
-      }
-    });
+          PgUtil.post(PROXY_FOR_TABLE, entity, okapiHeaders, vertxContext,
+            PostProxiesforResponse.class, asyncResultHandler);
+        }
+      });
```
julianladisch: Improve readability by using guard clauses. This will reduce the indentation level of the PgUtil.post, the main path of the execution flow:

    if (existsRes.failed()) {
      …
      return;
    }
    if (existsRes.result() == true) {
      …
      return;
    }
    PgUtil.post(…

src/main/java/org/folio/rest/impl/ProxiesForAPI.java
adamdickmeiss: nice

src/main/java/org/folio/rest/impl/ProxiesForAPI.java
adamdickmeiss: sure

src/main/java/org/folio/rest/impl/UserGroupAPI.java
```diff
@@ -40,323 +32,176 @@
...
-          entity.setId(id);
+    Map<String, String> okapiHeaders,
+    Handler<AsyncResult<Response>> asyncResultHandler,
+    Context vertxContext) {
+
+    String id = entity.getId();
+    if (id == null) {
+      id = UUID.randomUUID().toString();
+      entity.setId(id);
+    }
```
julianladisch: Generating randomUUID() can be handled by PgUtil.post:
 https://github.com/folio-org/raml-module-builder/blob/v25.0.0/domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java#L505

src/main/java/org/folio/rest/impl/UserGroupAPI.java
```diff
@@ -40,323 +32,176 @@
...
+    PostgresClientUtil.getInstance(vertxContext, okapiHeaders).save(
```
julianladisch: Consider using PgUtil.post

adamdickmeiss: That results in a regression.
Failed tests: 
  RestVerticleIT.test2Group:1737 Test failed

src/main/java/org/folio/rest/impl/UserGroupAPI.java
```diff
@@ -40,323 +32,176 @@
...
-        });
-      } catch (Exception e) {
-        log.error(e.getMessage(), e);
-        asyncResultHandler.handle(io.vertx.core.Future.succeededFuture(GetGroupsByGroupIdResponse
-          .respond500WithTextPlain(messages.getMessage(lang, MessageConsts.InternalServerError))));
-      }
-    });
+    Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
+
+    // PgUtil.getById .. returns 500 where the below returns 404
```
julianladisch: Returning 500 for an id string that isn't in valid UUID format is reasonable because the client most likely has mixed up something.
I suggest to change the unit test to both accept 404 and 500 for an invalid UUID.
And to use PgUtil.getById.

src/main/java/org/folio/rest/impl/UserGroupAPI.java
```diff
@@ -40,323 +32,176 @@
...
+    Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
+
+    Criterion criterion = new Criterion(
+      new Criteria()
+        .addField(ID_FIELD_NAME)
+        .setJSONB(false)
+        .setOperation("=")
+        .setValue("'" + groupId + "'"));
+    PostgresClient postgresClient = PostgresClientUtil.getInstance(vertxContext, okapiHeaders);
+    postgresClient.get(GROUP_TABLE, Usergroup.class, criterion, true, getReply -> {
```
julianladisch: If this is really needed it should be replaced by PgUtil.getById

However, PgUtil.deleteById automatically returns 404 if the records doesn't exists. Therefore we may skip this check.

If we were using a foreign key relationship (schema.json) between users and user groups we can simply invoke deleteById and it will fail properly if the user group is still in use, no manual check needed.

src/main/java/org/folio/rest/impl/UserGroupAPI.java
adamdickmeiss: I don't want to change behavior. If you pass something in as a client it's CLIENT ERROR (4XX).

src/main/java/org/folio/rest/impl/UserGroupAPI.java
adamdickmeiss: I'll need a commit from your to clarify. Or we skip this. Thanks.

src/main/java/org/folio/rest/impl/UserGroupAPI.java
julianladisch: This is caused by the "try to add a duplicate group" test.
PgUtil.post correctly returns 400 with a plain/text body. This complies with https://github.com/folio-org/raml/blob/raml1.0/rtypes/collection.raml#L38-L70

The RestVerticleIT.test2Group unit test expects a 422 application/json body:
https://github.com/folio-org/raml/blob/raml1.0/traits/validation.raml
The validation trait is optional for any post endpoint and is intended for additional tests that goes beyond what the regular collection.raml checks.

If you think that PgUtil.post should check for the optional 422 validation trait please file an issue against PgUtil.

Otherwise consider to either accept any return code (400 or 422), or remove the validation trait.

src/main/java/org/folio/rest/impl/UserGroupAPI.java
julianladisch: That's true, thanks for the clarification.

Opened https://issues.folio.org/browse/RMB-400 "PgUtil should return 400 (not 500) on invalid UUID id"

adamdickmeiss:
Thx. And now, unrelated to your changes, Docker failure!!

PR #120, opened 2019-06-17
Upgrade to RMB 25.0.1 MODUSERS-126

adamdickmeiss:


src/main/java/org/folio/rest/utils/ExpirationTool.java
```diff
@@ -123,11 +122,10 @@ public static void doExpiration(Vertx vertx, Context context) {
...
         PostgresClient pgClient = PostgresClient.getInstance(vertx, tenant);
-        pgClient.setIdField("id");
         Criteria idCrit = new Criteria();
         idCrit.addField(USER_ID_FIELD);
         idCrit.setOperation("=");
-        idCrit.setValue(user.getId());
+        idCrit.setVal(user.getId());
         Criterion criterion = new Criterion(idCrit);
 
         pgClient.update(TABLE_NAME_USERS, user, criterion, true, updateReply -> {
```
julianladisch: Can you drop Criteria/Criterion and directly use the method that takes the id parameter?

    pgClient.update(TABLE_NAME_USERS, user, user.getId(), updateReply -> {

src/main/java/org/folio/rest/impl/UsersAPI.java
```diff
@@ -512,7 +512,7 @@ private void getPG(PostgresClient postgresClient, User user, Handler<AsyncResult
    }else{
      Criterion c = new Criterion(
        new Criteria().addField(UserGroupAPI.ID_FIELD_NAME).setJSONB(false).
-       setOperation("=").setValue("'"+pgId+"'"));
+       setOperation("=").setVal(pgId));
      /** check if the patron group exists, if not, can not add the user **/
      postgresClient.get(
        UserGroupAPI.GROUP_TABLE, Usergroup.class, c, true, false, check -> {
```
julianladisch: Can you drop Criteria/Criteria and directly use a method that takes the pgId?

      postgresClient.getByIdAsString(UserGroupAPI.GROUP_TABLE, pgId, check -> {
         if (check.succeeded()) {
           if (check.result() == null) {
             handler.handle(io.vertx.core.Future.succeededFuture(0));
           } else {
             handler.handle(io.vertx.core.Future.succeededFuture(1));
           }
         }

julianladisch: Can you drop Criteria/Criteria and directly use a method that takes the pgId?

      postgresClient.getByIdAsString(UserGroupAPI.GROUP_TABLE, pgId, check -> {
         if (check.succeeded()) {
           if (check.result() == null) {
             handler.handle(io.vertx.core.Future.succeededFuture(0));
           } else {
             handler.handle(io.vertx.core.Future.succeededFuture(1));
           }
         }

src/main/java/org/folio/rest/utils/PostgresClientUtil.java
```diff
@@ -22,8 +22,7 @@ private PostgresClientUtil() {
    * @return the PostgresClient instance for the vertx and the tenantId
    */
   public static PostgresClient getInstance(Context vertxContext, Map<String, String> okapiHeaders) {
-    PostgresClient postgresClient = PostgresClient.getInstance(
+    return PostgresClient.getInstance(
         vertxContext.owner(), TenantTool.tenantId(okapiHeaders));
-    return postgresClient;
```
julianladisch: The change is ok, but I would rather prefer to drop PostgresClientUtil and use PgUtil.postgresClient instead.

src/main/java/org/folio/rest/impl/UsersAPI.java
```diff
@@ -506,34 +505,26 @@ else if(res == -1){
   */
  private void getPG(PostgresClient postgresClient, User user, Handler<AsyncResult<Integer>> handler) {
    String pgId = user.getPatronGroup();
-   if(pgId == null){
+   if (pgId == null){
```
kostyantyn-kh:
```suggestion
   if (pgId == null) {
```

adamdickmeiss:
Thanks for some reformatting. There's plenty of untested areas so reformatting .. indentation and if(-stuff is still around. I'll try to make a separate PR for that and see how SonarQube complains.

PR #121, opened 2019-06-18
Use PgUtil.postgresClient; reformat and SQ suggestions

adamdickmeiss:


src/main/java/org/folio/rest/impl/UsersAPI.java
```diff
@@ -499,45 +483,43 @@ public void putUsersByUserId(String userId,
...
+          if (t.getLocalizedMessage().contains("uuid")) {
+            retCode = 0;
+          }
+          handler.handle(io.vertx.core.Future.succeededFuture(retCode));
+        }
+      });
+    }
+  }
 
   private boolean checkForDuplicateAddressTypes(User user) {
```
julianladisch: Can this be static because it depends on the user parameter only?

src/main/java/org/folio/rest/impl/UsersAPI.java
```diff
@@ -499,45 +483,43 @@ public void putUsersByUserId(String userId,
...
     Map<String, Integer> countMap = new HashMap<>();
-    if (user.getPersonal() != null &&
-      user.getPersonal().getAddresses() != null) {
-      for(Address address : user.getPersonal().getAddresses()) {
+    if (user.getPersonal() != null
+      && user.getPersonal().getAddresses() != null) {
+      for (Address address : user.getPersonal().getAddresses()) {
```
julianladisch: Can this if clause been converted into a guard clause to reduce the indention of the following code and increase readability?

    if (user.getPersonal() == null || user.getPersonal().getAddresses() == null) {
      return false;
    }

src/main/java/org/folio/rest/impl/UsersAPI.java
```diff
@@ -499,45 +483,43 @@ public void putUsersByUserId(String userId,
...
+            retCode = 0;
+          }
+          handler.handle(io.vertx.core.Future.succeededFuture(retCode));
+        }
+      });
+    }
+  }
 
   private boolean checkForDuplicateAddressTypes(User user) {
     Map<String, Integer> countMap = new HashMap<>();
```
julianladisch: Can a `Set<String> found` been used instead?

src/main/java/org/folio/rest/impl/UsersAPI.java
```diff
@@ -499,45 +483,43 @@ public void putUsersByUserId(String userId,
...
+      for (Address address : user.getPersonal().getAddresses()) {
         String addressTypeId = address.getAddressTypeId();
         if (addressTypeId != null) {
           boolean found = false;
-          for(String key : countMap.keySet()) {
-            if (key.equals(addressTypeId)) {
-              Integer count = countMap.get(key);
-              count = count + 1;
-              countMap.put(key, count);
+          for (Entry<String,Integer> ent : countMap.entrySet()) {
```
julianladisch: Why do we need to iterate? Isn't this a hash?

src/main/java/org/folio/rest/impl/UsersAPI.java
```diff
@@ -499,45 +483,43 @@ public void putUsersByUserId(String userId,
...
+            if (ent.getKey().equals(addressTypeId)) {
+              countMap.put(ent.getKey(), ent.getValue() + 1);
```
julianladisch: This increments the value from (at least) 1 to (at least) 2. Can't we immediately return true in that case?

PR #123, opened 2019-07-05
MODUSERS-131: Update to RMB 26.1.0

julianladisch:
* Use latest PgUtil.post, PgUtil.getById, PgUtil.deleteById, PgUtil.put where
  possible, RMB 26.1.0 has fixed some HTTP status codes.
* Use foreign key users.patronGroup=groups.id
* Extract duplicate unit test code into fail and assertStatus methods.
* Remove postgres-embedded code, this is automatically handled by PostgresClient.


adamdickmeiss:
Note that there are a couple of trivial smells reported by SQ

PR #132, opened 2019-08-28
MODUSERS-142: Update raml submodule

julianladisch:
to make metadata.createdByUserId optional


julianladisch:
This also fixes https://issues.folio.org/browse/MODUSERS-143

PR #136, opened 2019-09-17
MODUSERS-146 - Integrate custom fields to mod-users

natali-zaitseva:
## Purpose
Due to https://issues.folio.org/browse/MODUSERS-146 we would like to introduce custom field for mod-users

## Approach
* added custom fields as a dependency
* modified Module Descriptor with an interface of the custom fields
* added custom fields table creation in the scripts section

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -117,6 +117,38 @@
       ]
     },
     {
+      "id": "custom-fields",
+      "version": "1.0",
+      "interfaceType" : "multiple",
+      "handlers": [
+        {
+          "methods": ["GET"],
+          "pathPattern": "/custom-fields",
+          "permissionsRequired": ["custom.fields.collection.get"]
+        },
+        {
+          "methods": ["POST"],
+          "pathPattern": "/custom-fields",
+          "permissionsRequired": ["custom.fields.item.post"]
+        },
+        {
+          "methods": ["GET"],
+          "pathPattern": "/custom-fields/{id}",
+          "permissionsRequired": ["custom.fields.item.get"]
+        },
+        {
+          "methods": ["PUT"],
+          "pathPattern": "/custom-fields/{id}",
+          "permissionsRequired": ["custom.fields.item.put"]
+        },
+        {
+          "methods": ["DELETE"],
+          "pathPattern": "/custom-fields/{id}",
+          "permissionsRequired": ["custom.fields.item.delete"]
+        }
+      ]
+    },
+    {
       "id": "_tenant",
       "version": "1.2",
       "interfaceType" : "system",
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -124,27 +124,27 @@
         {
           "methods": ["GET"],
           "pathPattern": "/custom-fields",
-          "permissionsRequired": ["custom.fields.collection.get"]
+          "permissionsRequired": ["user-settings.custom.fields.collection.get"]
         },
         {
           "methods": ["POST"],
           "pathPattern": "/custom-fields",
-          "permissionsRequired": ["custom.fields.item.post"]
+          "permissionsRequired": ["user-settings.custom.fields.item.post"]
         },
         {
           "methods": ["GET"],
           "pathPattern": "/custom-fields/{id}",
-          "permissionsRequired": ["custom.fields.item.get"]
+          "permissionsRequired": ["user-settings.custom.fields.item.get"]
         },
         {
           "methods": ["PUT"],
           "pathPattern": "/custom-fields/{id}",
-          "permissionsRequired": ["custom.fields.item.put"]
+          "permissionsRequired": ["user-settings.custom.fields.item.put"]
         },
         {
           "methods": ["DELETE"],
           "pathPattern": "/custom-fields/{id}",
-          "permissionsRequired": ["custom.fields.item.delete"]
+          "permissionsRequired": ["user-settings.custom.fields.item.delete"]
         }
       ]
     },
@@ -299,6 +299,43 @@
         "proxiesfor.item.put",
         "proxiesfor.item.delete"
       ]
+    },{
+      "permissionName": "user-settings.custom.fields.collection.get",
+      "displayName": "User Custom Fields - get collection",
+      "description": "Get User Custom Fields collection"
+    },
+    {
+      "permissionName": "user-settings.custom.fields.item.post",
+      "displayName": "User Custom Fields - create field",
+      "description": "Create User Custom Field"
+    },
+    {
+      "permissionName": "user-settings.custom.fields.item.get",
+      "displayName": "User Custom Fields - get field",
+      "description": "Get User Custom Field"
+    },
+    {
+      "permissionName": "user-settings.custom.fields.item.put",
+      "displayName": "User Custom Fields - modify field",
+      "description": "Modify User Custom Field"
+    },
+    {
+      "permissionName": "user-settings.custom.fields.item.delete",
+      "displayName": "User Custom Fields - delete field",
+      "description": "Delete User Custom Field"
+    },
+    {
+      "permissionName": "user-settings.custom.fields.all",
+      "displayName": "User Custom Fields module - all permissions",
+      "description": "Entire set of permissions needed to use the user custom fields",
+      "subPermissions": [
+        "user-settings.custom.fields.collection.get",
+        "user-settings.custom.fields.item.post",
+        "user-settings.custom.fields.item.get",
+        "user-settings.custom.fields.item.put",
+        "user-settings.custom.fields.item.delete"
+      ],
+      "visible": false
     }
   ],
   "metadata": {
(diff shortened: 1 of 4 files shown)


dmtkachenko:
> 
> 
> The UXPROD-33 key requirements https://docs.google.com/presentation/d/1zXvDSw-zJ36dKWv1C-W3-x8yOmKNu01c3LV0prJh1F8/edit#slide=id.g5d804cb54a_0_32
> say: "Custom fields are app+record type specific. (i.e. custom fields created in Settings > Users > is only available on the User record)"
> 
> I don't see how the current implementation is record type specific. How can I distinguish custom fields of the user records from custom fields of the usergroup records?
> 
> The interface name ModuleDescriptor is "custom-fields". The name must be specific for app+record type. For example "users-custom-fields" or "usergroups-custom-fields".
> 
> say: "Custom fields are app+record type specific. (i.e. custom fields created in Settings > Users > is only available on the User record)"
I don't see how the current implementation is record type specific. How can I distinguish custom fields of the user records from custom fields of the usergroup records?

It's record type specific at the level of custom-field data: there is "entityType" attribute to assign customer field to a specific record. Thus all user CFs can be separated from user-group CFs (can be done with appropriate cqlquery)
I don't see any reason in providing multiple endpoint (each per record type)

natali-zaitseva:
@julianladisch as for 

> Custom fields are app+record type specific. (i.e. custom fields created in Settings > Users > is only available on the User record)"

each module which needs to support custom fields will have a table defined for its schema, i.e `mod-users` will have an additional table called `custom_fields` to store module-specific custom fields definitions.

> I don't see how the current implementation is record type specific. How can I distinguish custom fields of the user records from custom fields of the usergroup records?

The record type is assumed to come from the UI(we currently do not have validation for the type) and in case of creation custom field for user record, we will specify property called `entityType` with the appropriate value --> `entityType` : `user` . 
The custom fields properties and description could be found [here](https://github.com/folio-org/mod-custom-fields/blob/master/ramls/customField.json)

and here is an example of the request 
Screen Shot 2019-09-17 at 5 44 29 PM


> The interface name ModuleDescriptor is "custom-fields". The name must be specific for app+record type. For example "users-custom-fields" or "usergroups-custom-fields".

I am not sure that we have a strong requirement for this. It is supposed to an interface that a lot of modules might use and it should be the same for all of them and the list of the modules will be available via the following call

Screen Shot 2019-09-17 at 5 46 27 PM

> Can you add one unit test for each of the 5 new endpoints: collection.get, item.post, item.get, item.put, item.delete?

will do.

descriptors/ModuleDescriptor-template.json
```diff
@@ -116,6 +116,38 @@
...
     },
+    {
+      "id": "custom-fields",
+      "version": "1.0",
+      "interfaceType" : "multiple",
+      "handlers": [
+        {
+          "methods": ["GET"],
+          "pathPattern": "/custom-fields",
+          "permissionsRequired": ["custom.fields.collection.get"]
```
dmtkachenko: Please rename all permissions to start with "user-settings.", like "user-settings.custom.fields.collection.get" instead of "custom.fields.collection.get".

For other modules we would need differently named permissions so we cannot use simply "custom.fields.collection.get" w/o a prefix.

natali-zaitseva: thanks, @dmtkachenko, renamed

src/test/java/org/folio/moduserstest/RestVerticleIT.java
```diff
@@ -2198,4 +2189,132 @@ private static String makeFakeJWT(String username, String id, String tenant) {
...
+
+  private Future<Void> assertCustomFieldValues(JsonObject result) {
+    Future<Void> future = Future.succeededFuture();
+    int totalRecords = result.getInteger("totalRecords");
+    if (totalRecords != 1) {
+      future.fail("Expected 1 record, got " + totalRecords);
+    }
+    JsonArray customFields = result.getJsonArray("customFields");
+    JsonObject customField = customFields.getJsonObject(0);
+    assertEquals(customField.getString("entityType"), "user");
```
julianladisch: The assert parameters are

    assertEquals(expected, actual);

or

    assertThat(actual, is(expected));

Therefore the correct parameter order is

    assertThat(customField.getString("entityType"), is("user"));

or

    assertEquals("user", customField.getString("entityType"));

descriptors/ModuleDescriptor-template.json
```diff
@@ -267,6 +299,43 @@
         "proxiesfor.item.put",
         "proxiesfor.item.delete"
       ]
+    },{
+      "permissionName": "user-settings.custom.fields.collection.get",
```
julianladisch: Can you rename `custom.fields` to `custom-fields` (similar to `user-settings`)?

descriptors/ModuleDescriptor-template.json
```diff
@@ -116,6 +116,38 @@
         }
       ]
     },
+    {
+      "id": "custom-fields",
```
julianladisch: Can you rename `custom-fields` to `users-custom-fields`?
Each module may have custom fields. The id of mod-inventory-storage will probably be `inventory-custom-fields`.

descriptors/ModuleDescriptor-template.json
```diff
@@ -116,6 +116,38 @@
...
+      "version": "1.0",
+      "interfaceType" : "multiple",
+      "handlers": [
+        {
+          "methods": ["GET"],
+          "pathPattern": "/custom-fields",
```
julianladisch: Can you rename `/custom-fields` to `/users-custom-fields`?
Each module may have custom fields. The pathPattern of mod-inventory-storage will probably be `/inventory-custom-fields`.
This will result in these URLs:
https://folio-snapshot-stable-okapi.aws.indexdata.com/users-custom-fields
https://folio-snapshot-stable-okapi.aws.indexdata.com/inventory-custom-fields

julianladisch:
Thank you for the explanation about the entityType!

descriptors/ModuleDescriptor-template.json
natali-zaitseva: we have defined api for custom fields in [custom fields module](https://github.com/folio-org/mod-custom-fields/blob/master/ramls/custom-fields.raml) and if I change the routing in `mod-users` to `/users-custom-fields` it will simply not work because there is no implementation for such endpoint

descriptors/ModuleDescriptor-template.json
natali-zaitseva: ok, will be changed

natali-zaitseva:
> Thank you, almost all issues have been resolved!
> The cleanup in RestVerticleIT.java is ok but you could have added a new CustomFieldsIT.java for your tests. You can leave your new tests in RestVerticleIT.java.
> The interface "id" a module provides must be different from the interface "id" of all other modules. It is not possible that both mod-users and mod-inventory-storage have the same interface `"id": "custom-fields"`.
> Same issue with pathPattern.

actually possible, according to the https://github.com/folio-org/okapi/blob/master/doc/guide.md#multiple-interfaces

descriptors/ModuleDescriptor-template.json
julianladisch: When Okapi gets a request for 
https://folio-snapshot-stable-okapi.aws.indexdata.com/custom-fields
how should Okapi know whether to route it to mod-users or mod-inventory-storage?
It looks as if https://github.com/folio-org/mod-custom-fields needs to be changed to allow each module having a different pathPattern.

natali-zaitseva: by the header `X-Okapi-Module-Id`. Please have a look on my first picture posted in this PR

natali-zaitseva: I have already tested it on local vagrant env and posted an image from the real case

descriptors/ModuleDescriptor-template.json
julianladisch: Because of `"interfaceType" : "multiple"` the rename is not needed: https://github.com/folio-org/okapi/blob/master/doc/guide.md#multiple-interfaces

descriptors/ModuleDescriptor-template.json
julianladisch: Yes, you are right. Sorry, I've never seen an implementation of "interfaceType": "multiple" before.

julianladisch:
Sorry, interface "id" and "pathPattern" can actually be the same when using "interfaceType": "multiple".

julianladisch:
Sorry for the confusion. There probably needs to be some documentation added about the multiple interface and the X-Okapi-Module-Id header in mod-custom-fields: https://issues.folio.org/browse/MODCFIELDS-12

src/test/java/org/folio/moduserstest/RestVerticleIT.java
natali-zaitseva: changed, thanks @julianladisch

natali-zaitseva:
@julianladisch thanks for creation issue. If there are no other comments related to this PR from your side could you merge the issue, please?

PR #137, opened 2019-09-25
MODUSERS-149: Add custom fields to user

andrii-paias:
## Purpose
Add option to save custom fields user

### DIFF
--- a/src/test/java/org/folio/moduserstest/RestVerticleIT.java
+++ b/src/test/java/org/folio/moduserstest/RestVerticleIT.java
@@ -236,6 +236,12 @@ public class RestVerticleIT {
     u.put("tags", tagobj);
   }
 
+  private static void addCustomFields(JsonObject u) {
+    JsonObject customFields = new JsonObject();
+    customFields.put("department", "math");
+    u.put("customFields", customFields);
+  }
+
   private Future<Void> postUser(TestContext context, boolean withUserName) {
     System.out.println("Creating a new user\n");
     Future<Void> future = Future.future();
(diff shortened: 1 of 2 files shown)


src/test/java/org/folio/moduserstest/RestVerticleIT.java
```diff
@@ -236,6 +236,12 @@ private static void addTags(JsonObject u) {
     u.put("tags", tagobj);
   }
 
+  private static void addCustomFields(JsonObject u) {
+    JsonObject customFields = new JsonObject();
+    customFields.put("department", "math");
```
dmtkachenko: 1.  _"department"_ -- is it a refId of CF or label or other?
2. do you see any differences in storing values for custom fields of different type:
- is it always going to be a key and string value or it might be a key and integer value (for instance for single-select field, an index of selected option could be stored instead of option itself)
- how would multi-select values look like in a record? if differently could you please add another test case

andrii-paias: 1. It is refId
2. Structure of custom field will depend on field type, fields with multiple values will be represented by array of strings,
I think it is ok to store select fields as Strings with value of selected option, we could store it as uuid of a particular option(in this case we would have to generate uuid for each option of custom field), but that doesn't seem necessary. Storing it as index seems brittle because list of options can be changed.
Added test for custom field with type of array

dmtkachenko:
> 
> 
> This change altered the definition of a user in the API, should have meant the interface version had a minor version increment to reflect the new property. Without that, the client has no way of asking specifically for this new API.

https://github.com/folio-org/mod-users/pull/147 created to address it

PR #138, opened 2019-09-25
MODUSERS-149: Add custom field validation

andrii-paias:
## Purpose
Add custom field validation

## Approach
Refactor post and put user endpoints to allow chaining Future from validateCustomFields with rest of the code in those endpoints.

This PR is difficult to read by looking at changes from the whole PR, so here's explanation for each commit
https://github.com/folio-org/mod-users/pull/138/commits/6a23d68701f6444fa88ca73e2cd4c4766931d633?w=1 - Unwrap code from vertxContext.runOnContext, calling runOnContext is equivalent to running code immediately, see "Learning" section below
https://github.com/folio-org/mod-users/pull/138/commits/3026aeaa60650c85fca831fa74aaa1e56ae8599d?w=1 - replace 
```
try
{...}
catch()
{...} 
```
with
```
Future.succeededFuture()
.compose(...)
.otherwise(...)
```
The second structure also handles thrown exceptions, but it also allows to process exceptions returned in the form of failed futures (exceptions thrown from asynchronous code)
https://github.com/folio-org/mod-users/pull/138/commits/319202599e03261b804431e954cc68c6174d6339?w=1 - replace 
```
 .setHandler(...) 
```
with 
```
.compose(...)
```
compose will be triggered only if future suceeded, and the case of exception will be handled by otherwise(...) block from previous commit

https://github.com/folio-org/mod-users/pull/138/commits/c74d1a8569d238d61a06ab3e81724680372fbb04?w=1 - remove try-catch from compose(...) block, because exceptions thrown from outer compose(...) blocks will be handled by otherwise(...) block.
Add another compose(...) block with call to ValidationServiceImpl#validateCustomFields. validateCustomFields will throw CustomFieldValidationException if validation fails, this exception is handled in otherwise(...)

https://github.com/folio-org/mod-users/pull/138/commits/a917165364a641634fa71b41caf9a5605ba145b6?w=1 - Change structure 
```
.compose(o -> {
    new ValidationServiceImpl().validateCustomFields(...)
     .compose(...)
     .compose(...)
}
)
.otherwise(...)
```
into structure
```
.compose(o -> new ValidationServiceImpl().validateCustomFields(...))
   .compose(...)
   .compose(...)
   .otherwise(...)
```
This way exceptions can be handled by outer .otherwise(...)
https://github.com/folio-org/mod-users/pull/138/commits/12b05281f0a8a4c79f07fb00d4a868b87876e7c3?w=1 - Unwrap method from runOnContext similarly to first commit. Add call to ValidationServiceImpl#validateCustomFields. Add exception handling for validateCustomFields.

https://github.com/folio-org/mod-users/pull/138/commits/08a859a6fef6660747a5bbb02d045ff74546a1e2?w=1 - Add unit tests for custom field validation 

https://github.com/folio-org/mod-users/pull/138/commits/366145014c76ee45dbd659d269a91ea3794be0da?w=1 - extract some blocks of code into separate methods to fix Sonar issues

https://github.com/folio-org/mod-users/pull/138/commits/149a8bd4d6eb5099e1b67f8f3915e565c4522d4a?w=1 - 
If future from checkAllAddressTypesValid is returned into outer
future chain then exception will be handled by otherwise clause,
so "if (checkRes.failed())" part is unnecessary

https://github.com/folio-org/mod-users/pull/138/commits/70673d2488602c2b6f555d3c425572f753ee1da4?w=1 -
Exceptions thrown from getPGAndValidate(...) will fail outer future chain
and will be processed by otherwise clause

## Learning
if vertxContext.runOnContext is called on the context of the current thread. (e.g. vertx.getOrCreateContext().runOnContext(...))
then it will execute code on the same context and it will be effectively the same as running code immediately, without using runOnContext. 
https://stackoverflow.com/questions/45348621/vertx-3-4-difference-between-running-code-immediately-and-context-runoncontext

### DIFF
--- a/src/main/java/org/folio/rest/impl/UsersAPI.java
+++ b/src/main/java/org/folio/rest/impl/UsersAPI.java
@@ -180,99 +180,91 @@ public class UsersAPI implements Users {
           Map<String, String> okapiHeaders,
           Handler<AsyncResult<Response>> asyncResultHandler,
           Context vertxContext) {
+    if(checkForDuplicateAddressTypes(entity)) {
+      asyncResultHandler.handle(Future.succeededFuture(
+          PostUsersResponse.respond400WithTextPlain(
+                  "Users are limited to one address per addresstype")));
+      return;
+    }
     try {
-      vertxContext.runOnContext( v -> {
-        if(checkForDuplicateAddressTypes(entity)) {
+      PostgresClient postgresClient = PgUtil.postgresClient(vertxContext, okapiHeaders);
+
+      checkAllAddressTypesValid(entity, vertxContext, postgresClient).setHandler(
+              checkRes -> {
+        if (checkRes.failed()) {
+          logger.error(checkRes.cause().getLocalizedMessage(), checkRes.cause());
           asyncResultHandler.handle(Future.succeededFuture(
-              PostUsersResponse.respond400WithTextPlain(
-                      "Users are limited to one address per addresstype")));
-          return;
-        }
-        try {
-          PostgresClient postgresClient = PgUtil.postgresClient(vertxContext, okapiHeaders);
+            PostUsersResponse.respond500WithTextPlain(
+              messages.getMessage(lang, MessageConsts.InternalServerError))));
+        } else if (Boolean.FALSE.equals(checkRes.result())) {
+          asyncResultHandler.handle(Future.succeededFuture(
+            PostUsersResponse.respond400WithTextPlain(
+                    "You cannot add addresses with non-existant address types")));
+        } else {
+          try {
+            getPG(postgresClient, entity, handler -> {
 
-          checkAllAddressTypesValid(entity, vertxContext, postgresClient).setHandler(
-                  checkRes -> {
-            if (checkRes.failed()) {
-              logger.error(checkRes.cause().getLocalizedMessage(), checkRes.cause());
-              asyncResultHandler.handle(Future.succeededFuture(
-                PostUsersResponse.respond500WithTextPlain(
-                  messages.getMessage(lang, MessageConsts.InternalServerError))));
-            } else if (Boolean.FALSE.equals(checkRes.result())) {
-              asyncResultHandler.handle(Future.succeededFuture(
-                PostUsersResponse.respond400WithTextPlain(
-                        "You cannot add addresses with non-existant address types")));
-            } else {
-              try {
-                getPG(postgresClient, entity, handler -> {
-
-                  int res = handler.result();
-                  if (res == 0) {
-                    String message = "Cannot add " +
-                        entity.getPatronGroup() +
-                        ". Patron group not found";
-                    logger.error(message);
-                    asyncResultHandler.handle(Future.succeededFuture(
-                        PostUsersResponse.respond400WithTextPlain(
-                            message)));
-                    return;
-                  } else if (res == -1) {
-                    asyncResultHandler.handle(Future.succeededFuture(
-                        PostUsersResponse
-                        .respond500WithTextPlain("")));
-                    return;
-                  } else {
-                    Date now = new Date();
-                    entity.setCreatedDate(now);
-                    entity.setUpdatedDate(now);
-                    PgUtil.post(TABLE_NAME_USERS, entity, okapiHeaders, vertxContext, PostUsersResponse.class, reply -> {
-                      if (isDuplicateIdError(reply)) {
-                        asyncResultHandler.handle(
-                            succeededFuture(PostUsersResponse.respond422WithApplicationJson(
-                                ValidationHelper.createValidationErrorMessage(
-                                    "id", entity.getId(),
-                                    "User with this id already exists"))));
-                        return;
-                      }
-                      if (isDuplicateUsernameError(reply)) {
-                        asyncResultHandler.handle(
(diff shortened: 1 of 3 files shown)


src/main/java/org/folio/rest/impl/UsersAPI.java
```diff
@@ -177,107 +181,82 @@ public void getUsers(String query, String orderBy,
...
-              PostUsersResponse.respond500WithTextPlain(
-              messages.getMessage(lang, MessageConsts.InternalServerError))));
+            logger.debug("Save successful");
+            asyncResultHandler.handle(reply);
+        });
     }
-  }
 
-  private boolean isDuplicateIdError(AsyncResult<Response> reply) {
+    private boolean isDuplicateIdError(AsyncResult<Response> reply) {
```
julianladisch: Can you fix the indentation?

src/main/java/org/folio/rest/impl/UsersAPI.java
```diff
@@ -322,94 +301,71 @@ public void putUsersByUserId(String userId,
...
+                .respond400WithTextPlain(
+                  "This barcode has already been taken")));
+            return;
+          }
+          logger.debug("Save successful");
+          asyncResultHandler.handle(reply);
+        });
+    }
+
+    /**
```
julianladisch: Can you fix the identation of the start of comment?

src/main/java/org/folio/rest/impl/UsersAPI.java
```diff
@@ -443,6 +399,27 @@ private void getPG(PostgresClient postgresClient, User user, Handler<AsyncResult
    }
  }
 
+ private void getAndValidatePatronGroup(PostgresClient postgresClient, User user, Handler<AsyncResult<Response>> asyncResultHandler, Handler<AsyncResult<Void>> onSuccess){
+     getPG(postgresClient, user, handler -> {
```
julianladisch: Can you fix the indentation?

src/main/java/org/folio/rest/impl/UsersAPI.java
```diff
@@ -322,94 +301,71 @@ public void putUsersByUserId(String userId,
...
+            asyncResultHandler.handle(
+              succeededFuture(PutUsersByUserIdResponse
+                .respond400WithTextPlain(
+                  "This barcode has already been taken")));
+            return;
+          }
+          logger.debug("Save successful");
+          asyncResultHandler.handle(reply);
+        });
+    }
```
julianladisch: Can you fix the indentation?

src/main/java/org/folio/rest/impl/UsersAPI.java
```diff
@@ -322,94 +301,71 @@ public void putUsersByUserId(String userId,
...
-    } catch (Exception e) {
-      logger.debug(e.getLocalizedMessage());
-      asyncResultHandler.handle(Future.succeededFuture(
-              PutUsersByUserIdResponse.respond500WithTextPlain(
-                      messages.getMessage(lang, MessageConsts.InternalServerError))));
-    }
   }
 
-  /**
+    private void updateUser(User entity, Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
```
julianladisch: Can you wrap this line?

src/main/java/org/folio/rest/impl/UsersAPI.java
```diff
@@ -443,6 +399,27 @@ private void getPG(PostgresClient postgresClient, User user, Handler<AsyncResult
...
+                             message)));
+         } else if (res == -1) {
+             asyncResultHandler.handle(Future.succeededFuture(
+                     PostUsersResponse
+                             .respond500WithTextPlain("")));
+         } else {
+             onSuccess.handle(Future.succeededFuture());
+         }
+     });
+ }
```
julianladisch: Can you fix the indentation?

src/main/java/org/folio/rest/impl/UsersAPI.java
```diff
@@ -443,6 +399,27 @@ private void getPG(PostgresClient postgresClient, User user, Handler<AsyncResult
    }
  }
 
+ private void getAndValidatePatronGroup(PostgresClient postgresClient, User user, Handler<AsyncResult<Response>> asyncResultHandler, Handler<AsyncResult<Void>> onSuccess){
```
julianladisch: Can you wrap this line and add some javadoc?

src/main/java/org/folio/rest/impl/UsersAPI.java
```diff
@@ -544,4 +521,13 @@ private boolean checkForDuplicateAddressTypes(User user) {
...
     return future;
   }
+
+  private Map<String, Object> getCustomFields(User entity) {
+    if(entity.getCustomFields()!= null){
+      return entity.getCustomFields().getAdditionalProperties();
+    }
+    else{
+      return Collections.emptyMap();
+    }
```
julianladisch: Using a guard clause increases readability because the main code is no longer wrapped in the else clause (reduced indentation).

      if (entity.getCustomFields() == null) {
        return Collections.emptyMap();
      }
      return entity.getCustomFields().getAdditionalProperties();

PR #142, opened 2019-10-17
MODUSERS-154: Fix unit tests related to mod-custom-fields

andrii-paias:
## Purpose
Fix unit tests related to mod-custom-fields


src/main/java/org/folio/rest/impl/UsersAPI.java
```diff
@@ -260,22 +262,42 @@ private void saveUser(User entity, Map<String, String> okapiHeaders,
     });
   }
 
-  private boolean isDuplicateIdError(AsyncResult<Response> reply) {
```
andrii-paias: These methods had to be changed because PgUtil class returns errors in different format now.
Additionally PUT /users endpoint doesn't have 422 error, so logic for PUT and POST endpoints is now different

PR #143, opened 2019-10-22
MODCFIELDS-25 - Custom Fields: Support fields order on POST /custom-f…

natali-zaitseva:
## Purpose
Updates for https://issues.folio.org/browse/MODCFIELDS-25

## Approach
* update custom fields payload for test


natali-zaitseva:
should be merged after https://github.com/folio-org/mod-custom-fields/pull/22

PR #151, opened 2020-01-08
MODUSERS-114: Log exception including stacktrace for streaming GET Users

julianladisch:
Pass JsonProcessingException to reply handler to return a failure.

Do not swallow the exception stacktrace, log it.

Use English or the tenant language or the language requested by the client, but
not the language of the server.

### DIFF
--- a/src/main/java/org/folio/rest/impl/UsersAPI.java
+++ b/src/main/java/org/folio/rest/impl/UsersAPI.java
@@ -190,7 +190,6 @@ public class UsersAPI implements Users {
               if (reply.succeeded()) {
                 response.write(String.format(JSON_USERS_FOOTER, cnt[0], cnt[0]));
               } else {
-                logger.error("Get users as a stream failed: " + reply.cause().getMessage(), reply.cause());
                 response.setStatusCode(500).setStatusMessage(reply.cause().getMessage());
               }
               response.end();
--- a/src/main/java/org/folio/rest/impl/UsersAPI.java
+++ b/src/main/java/org/folio/rest/impl/UsersAPI.java
@@ -2,6 +2,7 @@ package org.folio.rest.impl;
 
 import static io.vertx.core.Future.succeededFuture;
 
+import java.io.UncheckedIOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Date;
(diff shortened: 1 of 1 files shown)


src/main/java/org/folio/rest/impl/UsersAPI.java
```diff
@@ -183,13 +184,14 @@ public void getUsers(String query, String orderBy,
               try {
                 response.write(JSON_MAPPER.writeValueAsString(user), "UTF-8");
               } catch (JsonProcessingException e) {
-                logger.error("This error should never happen: " + e);
+                throw new UncheckedIOException(e);
```
hjiebsco: User json -> User in RMB, and then -> User json here again. How likely is the exception? But I am fine with re-throw.

src/main/java/org/folio/rest/impl/UsersAPI.java
```diff
@@ -183,13 +184,14 @@ public void getUsers(String query, String orderBy,
...
               }
             }, reply -> {
               if (reply.succeeded()) {
                 response.write(String.format(JSON_USERS_FOOTER, cnt[0], cnt[0]));
               } else {
-                response.setStatusCode(500).setStatusMessage(reply.cause().getLocalizedMessage());
+                logger.error("Get users as a stream failed: " + reply.cause().getMessage(), reply.cause());
```
hjiebsco: RMB captured and logged it as ERROR.

julianladisch: Yes, you are right. I'm removing the duplicate logging.

src/main/java/org/folio/rest/impl/UsersAPI.java
hjiebsco: Let's say the exception could happen (I do not know how), where are we going to handle it after the re-throw? Should there be another layer of try/catch to handle it and close the connection properly?

src/main/java/org/folio/rest/impl/UsersAPI.java
```diff
@@ -183,13 +184,13 @@ public void getUsers(String query, String orderBy,
...
               } catch (JsonProcessingException e) {
-                logger.error("This error should never happen: " + e);
+                throw new UncheckedIOException(e);
               }
             }, reply -> {
               if (reply.succeeded()) {
                 response.write(String.format(JSON_USERS_FOOTER, cnt[0], cnt[0]));
               } else {
-                response.setStatusCode(500).setStatusMessage(reply.cause().getLocalizedMessage());
+                response.setStatusCode(500).setStatusMessage(reply.cause().getMessage());
```
hjiebsco: I noticed that in streaming mode, once we start to write response body, we can not change response status code anymore. HTTP/2 can do that, but we are not there yet. So the best we can do for now is to write the error message to the response body. So this line needs a bit tweak.

julianladisch: Yes. I've created https://issues.folio.org/browse/MODUSERS-169

PR #153, opened 2020-01-10
Fix get user as stream return 200 on failure MODUSERS-169

adamdickmeiss:


src/main/java/org/folio/rest/impl/UsersAPI.java
```diff
@@ -174,24 +174,35 @@ public void getUsers(String query, String orderBy,
...
-                response.write(String.format(JSON_USERS_FOOTER, cnt[0], cnt[0]));
+              // at least first item could be written
+              if (cnt[0]++ > 0) {
+                response.write(",");
               } else {
+                response.write(JSON_USERS_HEADER);
+              }
+              response.write(s);
+            }, reply -> {
+              // can't return error if beyond first item
```
julianladisch: Create at least 2 user records.
Run this query that fetches all records: `/users?query=id==*&limit=200`
If the database fails on the second record this is the output:

```
HTTP/1.1 200 OK
content-type: application/json
transfer-encoding: chunked

{
	"users": [
    {"username": "fred",  ... }
  ],
  "totalRecords": 1,
  "resultInfo": {
    "totalRecords": 1,
    "facets": [],
    "diagnostics": []
  }
}
```
totalRecords is wrong.
The 200 status code is wrong, the interface definition guarantees 200 on success and 500 on failure: https://s3.amazonaws.com/foliodocs/api/mod-users/users.html#users_get
The result is exactly the same as for the query that matches a single record only:
`/users?query=username==fred&limit=200`

adamdickmeiss: With HTTP and streaming there's no way we can signal error in the stream itself.. (of course we if we buffer it all up, we can do, but then it's no longer streaming). Therefore the only option is to assume that items CAN be fetched and that we at least for the initial query signal error. As soon as items flow down we're bound to the status code.

PR #157, opened 2020-01-28
MODUSERS-179: Make PUT /custom-fields endpoint available

andrii-paias:
## Purpose
Make PUT /custom-fields endpoint available

## Approach
Update version of custom fields to the newest SNAPSHOT version
Add endpoint to ModuleDescriptor
Add permission

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -30,7 +30,7 @@
     <dependency>
       <groupId>org.folio</groupId>
       <artifactId>mod-custom-fields</artifactId>
-      <version>1.0.3</version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -30,7 +30,7 @@
     <dependency>
       <groupId>org.folio</groupId>
       <artifactId>mod-custom-fields</artifactId>
-      <version>1.0.3</version>
+      <version>1.0.4-SNAPSHOT</version>
```
dmtkachenko: dependency to a snapshot is not permitted
please release the library version and use it here

PR #159, opened 2020-02-03
MODUSERS-175 update db schema, MODUSERS-177 create API endpoints

roman-barannyk:
## Purpose
Resolves [MODUSERS-175](https://issues.folio.org/browse/MODUSERS-175), [MODUSERS-177](https://issues.folio.org/browse/MODUSERS-177)

- Create DB table to store patron block limit information
- Create CRUD API endpoints for patron block conditions

## Approach 
- Update db schema
- Test that patron-block-limits table has been created locally
- Create .raml file and generate API
- Create CRUD implementation
- Add tests

## Links
[MODUSERS-175](https://issues.folio.org/browse/MODUSERS-175)
[MODUSERS-177](https://issues.folio.org/browse/MODUSERS-177)


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -178,6 +178,10 @@
           "tOps" : "ADD"
         }
       ]
+    },
+    {
+      "tableName" : "patron_block_limits",
+      "withMetadata" : true
```
julianladisch: Can you add "fromModuleVersion"?

julianladisch:
GET and PUT API endpoints are needed. If there is no other issue for this it should be implemented in this issue.
INSERT and DELETE API endpoints will be handled by triggers implemented by MODUSERS-178 and MODUSERS-180

roman-barannyk:
> Java code for CRUD is missing.

added java code with CRUD implementation

src/main/resources/templates/db_scripts/schema.json
roman-barannyk: > Can you add "fromModuleVersion"?

added

ramls/patron-block-limit.json
```diff
@@ -0,0 +1,36 @@
...
+      "type": "string",
+      "description": "A UUID identifying the patron group",
+      "$ref": "raml-util/schemas/uuid.schema"
+    },
+    "conditionId": {
+      "type": "string",
+      "description": "A UUID identifying the condition",
+      "$ref": "raml-util/schemas/uuid.schema"
+    },
+    "limit": {
```
OleksandrVidinieiev: Maybe we should choose a different name for this field, e.g. `value`? In terms of this feature, a limit is an aggregate of referenced entities' IDs and the limit's **value**.

ramls/patron-block-limits.json
```diff
@@ -0,0 +1,23 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Collection of available patron block limits",
+  "type": "object",
+  "properties": {
+    "patronBlockLimits": {
+      "description": "List of of available patron block limits",
```
OleksandrVidinieiev:
```suggestion
      "description": "List of available patron block limits",
```

ramls/patron-block-limits.raml
```diff
@@ -0,0 +1,48 @@
+#%RAML 1.0
+title: Patron Block Limits
+version: v0.1
+baseUri: http://github.com/org/folio/mod-users
+
+documentation:
+  - title: mod-users Patron Block Limits API
+    content: This documents the API calls that can be made to query and manage patron blocks limits
```
OleksandrVidinieiev:
```suggestion
    content: This documents the API calls that can be made to query and manage patron block limits
```

src/main/java/org/folio/rest/impl/PatronBlockLimitsAPI.java
```diff
@@ -0,0 +1,118 @@
...
+  public void deletePatronBlockLimitsByPatronBlockLimitId(String patronBlockLimitId,
+    String lang, Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler,
+    Context vertxContext) {
+
+    PgUtil.deleteById(PATRON_BLOCK_LIMITS, patronBlockLimitId, okapiHeaders,
+      vertxContext, DeletePatronBlockLimitsByPatronBlockLimitIdResponse.class,
+      asyncResultHandler);
+  }
+
+  private Errors isEntityValid(PatronBlockLimit entity) {
```
OleksandrVidinieiev: Personally, if a method's name starts with `is...`, I expect it to return a boolean. How about renaming it to something like `validateEntity`?

src/main/java/org/folio/rest/impl/PatronBlockLimitsAPI.java
```diff
@@ -0,0 +1,118 @@
...
+
+    return MAX_OUTSTANDING_FEEFINE_BALANCE_ID.equals(entity.getConditionId())
+      ? validateRangeForDoubleValueType(entity)
+      : validateRangeForIntegerValueType(entity);
+  }
+
+  private Errors validateRangeForDoubleValueType(PatronBlockLimit entity) {
+    Double limit = entity.getLimit();
+    if (limit > 0.1 && limit < 9999.99) {
```
OleksandrVidinieiev: Description says:
> "Maximum outstanding fee/fine balance" may be blank or a currency value between 0.01 and 9999.99.

I sounds to me as `limit >= 0.1 && limit <= 9999.99`.

src/main/java/org/folio/rest/impl/PatronBlockLimitsAPI.java
```diff
@@ -0,0 +1,118 @@
...
+      return null;
+    }
+    return createValidationErrorMessage("limit", entity.getLimit().toString(),
+      "A maximum balance of 0 will result in all patrons in this group being blocked; " +
```
OleksandrVidinieiev: Is this message still relevant if the limit value is 0.001 or 10000.00?

src/main/java/org/folio/rest/impl/PatronBlockLimitsAPI.java
```diff
@@ -0,0 +1,123 @@
...
+    String lang, Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler,
+    Context vertxContext) {
+
+    PgUtil.deleteById(PATRON_BLOCK_LIMITS, patronBlockLimitId, okapiHeaders,
+      vertxContext, DeletePatronBlockLimitsByPatronBlockLimitIdResponse.class,
+      asyncResultHandler);
+  }
+
+  private Errors isEntityValid(PatronBlockLimit entity) {
+
```
OleksandrVidinieiev: How about extracting a local variable for `entity.getLimit()` here and passing it to `validateRangeFor*ValueType()` methods? This will save us 4  `entity.getLimit()` calls.

ramls/patron-block-limit.json
roman-barannyk: renamed

ramls/patron-block-limits.json
roman-barannyk: fixed

ramls/patron-block-limits.raml
roman-barannyk: updated

src/main/java/org/folio/rest/impl/PatronBlockLimitsAPI.java
roman-barannyk: thanks, updated

src/main/java/org/folio/rest/impl/PatronBlockLimitsAPI.java
roman-barannyk: Thanks, validation has been updated and this test scenario covered

src/main/java/org/folio/rest/impl/PatronBlockLimitsAPI.java
roman-barannyk: done

src/main/java/org/folio/rest/impl/PatronBlockLimitsAPI.java
roman-barannyk: It was done with the same code style as in another modules, for example: `org.folio.rest.impl.FixedDueDateSchedulesAPI#postFixedDueDateScheduleStorageFixedDueDateSchedules` (mod-circulation-storage)
Also from my perspective the method named as `validateEntity` should return void type or throw an exception, but it our case it returns Errors

src/main/java/org/folio/rest/impl/PatronBlockLimitsAPI.java
```diff
@@ -0,0 +1,129 @@
...
+        "A maximum balance of 0 will result in all patrons in this group being blocked; " +
+          "to skip this limit, leave value set to blank");
+    }
+    return createValidationErrorMessage(VALUE_FIELD, limit.toString(),
+      "Must be blank or a number from 0.01 to 9999.99");
+  }
+
+  private Errors validateRangeForIntegerValueType(Double limit) {
+    boolean isInt = limit % 1 == 0;
+    if (limit == 0 || (isInt && limit > 0 && limit < 999999)) {
```
OleksandrVidinieiev:
```suggestion
    if (isInt && limit >= 0 && limit <= 999999) {
```

OleksandrVidinieiev:
```suggestion
    if (isInt && limit >= 0 && limit <= 999999) {
```

src/main/java/org/folio/rest/impl/PatronBlockLimitsAPI.java
```diff
@@ -0,0 +1,129 @@
...
+      return null;
+    }
+    return createValidationErrorMessage(VALUE_FIELD, limit.toString(),
+      "Must be blank or a number from 0 to 999999");
```
OleksandrVidinieiev:
```suggestion
      "Must be blank or an integer from 0 to 999999");
```

src/main/java/org/folio/rest/impl/PatronBlockLimitsAPI.java
roman-barannyk: done

roman-barannyk: done

src/main/java/org/folio/rest/impl/PatronBlockLimitsAPI.java
roman-barannyk: done

src/main/java/org/folio/rest/impl/PatronBlockLimitsAPI.java
julianladisch: I agree with Oleksandr.
The name in mod-circulation-storage is misleading, too, don't copy that naming scheme.
The method name should be `validateEntity` or even `validate` because `validate...` does not indicate the return type.

julianladisch: There already are two methods that start with `validate` and return `Errors`:
`Errors validateRangeForDoubleValueType`
`Errors validateRangeForIntegerValueType`

roman-barannyk: OK, renamed.
I have renamed the similar method in PatronBlockConditionsAPI as well.

OleksandrVidinieiev:
@julianladisch, I left some feedback to your comments in the stories.

PR #164, opened 2020-02-25
[MODUSERS-180] Maintain Patron Blocks Limits Table entries: patron group deletion

OleksandrVidinieiev:
**Purpose**
When a patron group is deleted, delete all patron block limits assosiated with it.

This PR resolves [MODUSERS-180](https://issues.folio.org/browse/MODUSERS-180).

### DIFF
--- a/src/test/java/org/folio/moduserstest/GroupIT.java
+++ b/src/test/java/org/folio/moduserstest/GroupIT.java
@@ -134,155 +134,6 @@ public class GroupIT {
   }
 
   @Test
-  public void deletingGroupDeletesItsPatronBlockLimits(TestContext context) throws Exception {
-    /*
-      create a new patron group
-     */
-    final String newGroupId = "1de95200-72e4-4967-bdf8-257fb7559888";
-
-    JsonObject group = new JsonObject()
-      .put("id", newGroupId)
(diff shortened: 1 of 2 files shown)


src/test/java/org/folio/moduserstest/GroupIT.java
```diff
@@ -126,6 +132,71 @@ public static void tearDown() {
     future.join();
   }
 
+  @Test
+  public void deletingGroupDeletesItsPatronBlockLimits(TestContext context) throws Exception {
+    /*
+      create a new patron group
+     */
```
alexanderkurash: Can these be a one line //-style comments?

OleksandrVidinieiev: Sure, but IMHO block comments here offer better visual separation of the test into logical steps.

alexanderkurash: You can keep them, but I personally would just add empty lines for visual separation. Maybe it's just my prejudice agains /**/-style comments, but they don't nest and sometimes it's causing inconvenience. For example, if you try now to comment out this whole method with /**/ there'll be errors.

PR #168, opened 2020-05-14
MODUSERS-193 move patron block functionality to mod-patron-blocks

roman-barannyk:
## Purpose 
Remove patron block conditions and limits functionality due to moving it to mod-patron-blocks

Resolves [MODUSERS-193](https://issues.folio.org/browse/MODUSERS-193)

## Approach
- remove DB schema;
- remove init data;
- remove endpoints;


descriptors/ModuleDescriptor-template.json
```diff
@@ -168,60 +168,6 @@
           "pathPattern" : "/_/tenant"
         }
       ]
-    },
-    {
-      "id": "patron-block-conditions",
-      "version": "0.1",
-      "handlers": [
-        {
-          "methods": ["GET"],
```
alexanderkurash, line 177: Should we update module version when we remove interfaces from it?

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -178,16 +178,6 @@
...
-    },
-    {
-      "tableName" : "patron_block_conditions",
-      "fromModuleVersion" : "mod-users-16.0.1",
-      "withMetadata" : true
-    },
-    {
-      "tableName" : "patron_block_limits",
-      "fromModuleVersion" : "mod-users-16.0.1",
-      "withMetadata" : true
```
alexanderkurash, line 190: We should also delete tables

descriptors/ModuleDescriptor-template.json
roman-barannyk: Updated version of module to 16.2.1

roman-barannyk: Updated version of module to 16.2.1

src/main/resources/templates/db_scripts/schema.json
roman-barannyk: done

PR #169, opened 2020-06-01
Modusers 195

jeremythuff:
Resolves [MODUSERS-195](https://issues.folio.org/browse/MODUSERS-195)


pom.xml
```diff
@@ -30,7 +30,7 @@
     <dependency>
       <groupId>org.folio</groupId>
       <artifactId>mod-custom-fields</artifactId>
-      <version>1.3.0</version>
+      <version>1.3.1-SNAPSHOT</version>
```
bohdan-suprun, line 33: We don't usually add snapshot versions

jeremythuff: I agree, this was at the suggestion of the custom field folks, and defiantly was not working without it. Should this update hold until custom-fields is released?

bohdan-suprun: I'm fine to leave snapshot if it is updated to released version once available. The module can not be released util we have the lib released, I think.

marcjohnson-kint: @jeremythuff Please can you raise an issue for updating this dependency to a release version and add it as a blocker for releasing mod-users.

src/test/java/org/folio/moduserstest/GroupIT.java
```diff
@@ -363,8 +356,8 @@ public void test2Group(TestContext context) throws Exception {
         + "\nStatus - " + addExpiredUserResponse.code + " at "
         + System.currentTimeMillis() + " for " + addUserURL + " (addExpiredUser)");
       context.assertEquals(addExpiredUserResponse.code, 201);
-      CompletableFuture<Void> getExpirationCF = new CompletableFuture();
-      ExpirationTool.doExpirationForTenant(RestITSupport.vertx(), RestITSupport.context(), "diku").setHandler(
+      CompletableFuture<Void> getExpirationCF = new CompletableFuture<Void>();
+      ExpirationTool.doExpirationForTenant(RestITSupport.vertx(), RestITSupport.context(), "diku").onComplete(
         res -> getExpirationCF.complete(null));
```
marcjohnson-kint, line 361:
```suggestion
      ExpirationTool.doExpirationForTenant(RestITSupport.vertx(), RestITSupport.context(), "diku")
        .onComplete(res -> getExpirationCF.complete(null));
```

marcjohnson-kint:
@jeremythuff Do you have time to add a test for expiration of users, which is where most of the low coverage warning is coming from?

PR #185, opened 2020-09-10
MODUSERS-221: Fix ExpirationTool nspname LIKE and error logging

julianladisch:
`nspname LIKE '%_mod_users'`
matches diku_my_mod_users and diku_modmusers.
This is wrong.
Use `nspname ~ '^[^_]+_mod_users$'` instead.

Errors are logged using logger.info. Use logger.error instead.

Log tenant.

Remove unused Context parameter.


hjiebsco (approved):
Should we consider to adjust some Expiration log level from info to debug to reduce mod-users log output/noise?

julianladisch:
Yes, done.

PR #198, opened 2020-10-20
MODUSERS-234: Add property expiration offset to usergroup

rchr:
Refs https://issues.folio.org/browse/MODUSERS-234

In order to implement [UIU-1908](https://issues.folio.org/browse/UIU-1908) we need a property describing the default expiration offset in days for a patron group.
This PR adds the optional property _expirationOffset_ to the _usergroup_ schema.

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -4,7 +4,7 @@
   "provides": [
     {
       "id": "users",
-      "version": "15.2",
+      "version": "15.3",
       "handlers" : [
         {
           "methods": [ "GET" ],
--- a/ramls/groups.raml
+++ b/ramls/groups.raml
@@ -1,6 +1,6 @@
 #%RAML 1.0
 title: User Groups
-version: v1.0.1
+version: v1.1.0
 baseUri: http://github.com/org/folio/mod-users
 
 documentation:
--- a/ramls/groups.raml
+++ b/ramls/groups.raml
@@ -1,6 +1,6 @@
 #%RAML 1.0
 title: User Groups
-version: v1.1.0
+version: v15.3
 baseUri: http://github.com/org/folio/mod-users
 
 documentation:
--- a/ramls/usergroup.json
+++ b/ramls/usergroup.json
@@ -17,7 +17,7 @@
       "type": "string"
     },
     "expirationOffsetInDays": {
-      "description": "The standard expiration offset in days of this group",
+      "description": "The default period in days after which a newly created user that belongs to this group will expire",
       "type": "integer"
     },
     "metadata": {
--- a/ramls/usergroup.json
+++ b/ramls/usergroup.json
@@ -16,7 +16,7 @@
       "description": "A UUID identifying this group",
       "type": "string"
     },
-    "expirationOffset": {
+    "expirationOffsetInDays": {
       "description": "The standard expiration offset in days of this group",
       "type": "integer"
     },
--- a/ramls/usergroup.json
+++ b/ramls/usergroup.json
@@ -16,6 +16,10 @@
       "description": "A UUID identifying this group",
       "type": "string"
     },
+    "expirationOffset": {
+      "description": "The standard expiration offset in days of this group",
+      "type": "integer"
+    },
     "metadata": {
       "$ref": "raml-util/schemas/metadata.schema",
       "readonly": true
--- a/reference-data/groups/faculty.json
+++ b/reference-data/groups/faculty.json
@@ -1,5 +1,6 @@
 {
   "id" : "503a81cd-6c26-400f-b620-14c08943697c",
   "group":"faculty",
-  "desc":"Faculty Member"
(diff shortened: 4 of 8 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -4,7 +4,7 @@
   "provides": [
     {
       "id": "users",
-      "version": "15.2",
+      "version": "15.3",
```
marcjohnson-kint, line 7: Please also change the version in the RAML file for this interface

ramls/usergroup.json
```diff
@@ -16,6 +16,10 @@
       "description": "A UUID identifying this group",
       "type": "string"
     },
+    "expirationOffset": {
```
marcjohnson-kint, line 19: Previously, periods of time without any units have caused confusion.

If this is always going to be measured in days, please make that explicit in the name. 

```suggestion
    "expirationOffsetInDays": {
```

reference-data/groups/faculty.json
```diff
@@ -1,5 +1,6 @@
 {
   "id" : "503a81cd-6c26-400f-b620-14c08943697c",
   "group":"faculty",
-  "desc":"Faculty Member"
+  "desc":"Faculty Member",
+  "expirationOffset": 365
```
marcjohnson-kint, line 5: Should the reference groups have these values in existing systems?

If so, please write an upgrade script for the existing records.

marcjohnson-kint (commented):
It would be really nice if there were better tests for groups. I understand that is likely out of scope of this work.

descriptors/ModuleDescriptor-template.json
rchr: You're right. I've done that!

ramls/usergroup.json
rchr: Makes absolute sense. I changed that.

reference-data/groups/faculty.json
rchr: No, as the value is optional, there is no need for existing systems to have these values.
I added them to the reference-data to show usage in reference-data resp sample-data.

ramls/groups.raml
```diff
@@ -1,6 +1,6 @@
 #%RAML 1.0
 title: User Groups
-version: v1.0.1
+version: v1.1.0
```
marcjohnson-kint, line 3: Interface versions only have a a major and minor, please can this change this. It should also match the version in the module descriptor.

```suggestion
version: v15.3
```

ramls/usergroup.json
```diff
@@ -16,6 +16,10 @@
       "description": "A UUID identifying this group",
       "type": "string"
     },
+    "expirationOffsetInDays": {
+      "description": "The standard expiration offset in days of this group",
```
marcjohnson-kint, line 20: What impact does this value have, e.g. do groups expire?

rchr: This will be work of another story [UIU-1907](https://issues.folio.org/browse/UIU-1907).

In short, if expirationOffsetInDays is set for a user group, the expiration date field is populated by the current date plus the given offset in the frontend. If the selected patron does not have a defined offset, everything will work as it currently does.

marcjohnson-kint: > the expiration date field is populated by the current date plus the given offset in the frontend. If the selected patron does not have a defined offset, everything will work as it currently does.

Is that the expiration date for the user (not the group)?

If so, could it be helpful for folks using this documentation to understand this property affects the expiration date for users within this group?

rchr: > Is that the expiration date for the user (not the group)?

Yes, this will be expiration date for the user, based on the user's group. If the value is defined for the user's group, the frontend will pre-fill the corresponding field in the front-end based on this value. However, the person which is editing the user is able to change the pre-filled value.

> If so, could it be helpful for folks using this documentation to understand this property affects the expiration date for users within this group?

Yes, I may have to change the description to something like the following: "The standard expiration offset in days of a user within this group". Does that answer your question?
And I am not sure, if the description should specify that the value is used by the frontend?

marcjohnson-kint: > The standard expiration offset in days of a user within this group

By standard offset do you mean the default period after which a newly created user will expire?

rchr: Yes, this is the default period in days after which a newly created user will expire.

julianladisch: Can you change the description to clarify the intended usage?
`The default period in days after which a newly created user that belongs to this group will expire.`
We should not mention how creates new users, this is implementation and institution dependent. Some might be created in the front-end, some might be created by some back-end batch process, some might be created by some external software (central campus registry).

rchr: @julianladisch I updated the description as you suggested. It's a more meaningful description.

And you are right. New user may be created by other systems than the front-end.

marcjohnson-kint (approved):

julianladisch (approved):

PR #203, opened 2021-01-26
MODUSERS-242: Upgrade to RAML Module Builder 32.x

K-Felk:
Most changes are explained in the RMB upgrade guide:

github.com/folio-org/raml-module-builder/blob/master/doc/upgrading.md

In addition, tenant setup seems to have changed from a synchronous to an 
asynchronous operation.  Tests in this module that required a tenant to 
operate were written with synchronous setup in mind-thus, when the 
upgrade was applied, they all started failing.  All tenant initialization 
code had to be re-written to allow for asynchronous operation, which accounts 
for most of the changes made to tests.

The tenant ref API also had to be reworked-my understanding from the 
RMB folks is that overriding loadData has been the preferred method of 
loading reference and sample data to test tenants for a while now.

Refs: https://issues.folio.org/browse/MODUSERS-242


src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -1,48 +1,41 @@
...
+import org.folio.rest.jaxrs.model.TenantAttributes;
+import org.folio.rest.tools.utils.TenantLoading;
 
 public class TenantRefAPI extends TenantAPI {
- private static final Logger log = LoggerFactory.getLogger(TenantRefAPI.class);
 
+  private static final String SAMPLE_LEAD = "sample-data";
+  private static final String SAMPLE_KEY = "loadSample";
+  private static final String REFERENCE_KEY = "loadReference";
+  private static final String REFERENCE_LEAD = "ref-data";
```
marcjohnson-kint, line 18: If these constants are only used in one place, why do they need to be defined in the scope of the entire class?

Personally, I find it easier to comprehend code when I need to navigate around less. I would prefer these were either moved to the only method they are used by or inlined.

src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -1,48 +1,41 @@
...
+          TenantLoading tl = new TenantLoading();
+          tl.withKey(REFERENCE_KEY).withLead(REFERENCE_LEAD);
+          tl.withIdContent();
+          tl.add("groups");
+          tl.withIdContent();
+          tl.add("addresstypes");
+          tl.withKey(SAMPLE_KEY).withLead(SAMPLE_LEAD);
+          tl.withIdContent();
+          tl.add("users");
+          return tl.perform(attributes, headers, vertxContext, superRecordsLoaded);
```
marcjohnson-kint, line 38: Personally, I find it easier to follow code when it spaced out a bit. In particular for methods that have sufficient parameters to have a line continuation, I think a blank line between the signature and the implementation helps to identify them easier.

```suggestion

    return super.loadData(attributes, tenantId, headers, vertxContext)
        .compose(superRecordsLoaded -> {
          log.info("loading data to tenant");
          
          TenantLoading tl = new TenantLoading();
          
          tl.withKey(REFERENCE_KEY).withLead(REFERENCE_LEAD);
          tl.withIdContent();
          tl.add("groups");
          tl.withIdContent();
          tl.add("addresstypes");
          
          tl.withKey(SAMPLE_KEY).withLead(SAMPLE_LEAD);
          tl.withIdContent();
          tl.add("users");
          
          return tl.perform(attributes, headers, vertxContext, superRecordsLoaded);
```

src/test/java/org/folio/moduserstest/CustomFieldIT.java
```diff
@@ -85,57 +86,40 @@
     "\"order\": 1 }";
 
 
+  private static Vertx vertx;
+  static int port;
```
marcjohnson-kint, line 90: Is port used in any other method than `setup`?

src/test/java/org/folio/moduserstest/GroupIT.java
```diff
@@ -66,62 +67,44 @@
   private static final String fooGroupData = "{\"group\": \"librarianFOO\",\"desc\": \"yet another basic lib group\", \"expirationOffsetInDays\": 365}";
   private static final String barGroupData = "{\"group\": \"librarianBAR\",\"desc\": \"and yet another basic lib group\"}";
 
-  private static final Logger log = LoggerFactory.getLogger(GroupIT.class);
+  private static final Logger log = LogManager.getLogger(GroupIT.class);
 
   private static int userInc = 0;
 
+  private static Vertx vertx;
+  static int port;
```
marcjohnson-kint, line 75: Is port used anywhere other than the `setup` method?

src/test/java/org/folio/moduserstest/RestVerticleIT.java
```diff
@@ -89,61 +91,43 @@
     .put("requestForSponsor", "Yes")
     .put("notificationsTo", "Proxy")
     .put("accrueTo", "Sponsor");
-
-
+  
+  private static Vertx vertx;
+  static int port;
```
marcjohnson-kint, line 96: Is port used anywhere else other than the `setup` method?

src/test/java/org/folio/rest/impl/UsersAPIIT.java
```diff
@@ -91,20 +101,53 @@ static TenantAttributes tenantAttributes() {
...
+
+    waitForTenantInit(id);
+  }
+
+  //restassured has no way of doing asynchronous checking for tenant initialization
+  //without introducing more dependencies to support that.  Thus, a custom wait
+  //method to do this is neccessary.
+
+  private static void waitForTenantInit(String id) {
+
```
marcjohnson-kint, line 118: Personally, I don't think this blank line aids readability, I would remove it

src/test/java/org/folio/rest/impl/UsersAPIIT.java
```diff
@@ -91,20 +101,53 @@ static TenantAttributes tenantAttributes() {
...
+      }
+      tenantSetupDone = given().when().get("/_/tenant/" + id).then().statusCode(200).extract().path("complete");
+      tries++;
+    }
+    
+    assertEquals(tenantSetupDone, true);
+
+    //if something went wrong internally with client setup,
+    //there will be an error attribute in the response body
+    given().when().get("/_/tenant/" + id).then().statusCode(200).body("$", not(hasKey("error")));
```
marcjohnson-kint, line 136: In other modules (like mod-circulation) we have used libraries like Awaitility in order to avoid having to use loops and thread.sleep.

I think it could be worth doing that here as well.

src/test/java/org/folio/rest/utils/TenantInit.java
```diff
@@ -0,0 +1,57 @@
+package org.folio.rest.utils;
+
+import io.vertx.core.Future;
+import io.vertx.core.Promise;
+import io.vertx.core.json.JsonObject;
+import org.folio.rest.client.TenantClient;
+import org.folio.rest.jaxrs.model.TenantAttributes;
+
+public class TenantInit {
+
```
marcjohnson-kint, line 10: Personally, I don't think this blank line aids readability, I would remove it

src/test/java/org/folio/rest/impl/UsersAPIIT.java
K-Felk: The only reason I did not do this is because it means introducing a new dependency into the module just to support this test-I was not sure that Would Be Looked Favorably Upon.  If that's fine, then I'm happy to use it-it certainly makes this task easier.

marcjohnson-kint: I understand the caution over introducing new dependencies. FOLIO doesn't really have a clear stance on that.

In this case, I think it is preferable to using `thread.sleep`, the library is used elsewhere and it is only for testing.

NEWS.md
```diff
@@ -1,3 +1,7 @@
+## 17.3.0 2021-01-05
```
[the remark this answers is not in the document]
K-Felk: Mea culpa.  Changed to IN-PROGRESS.

NEWS.md
```diff
@@ -1,3 +1,7 @@
+## 17.3.0 IN-PROGRESS
+
+* Updated Vert.x to version 4 (MODUSERS-242)
+
 ## 17.2.0 2020-10-06
```
adamdickmeiss, line 5: Please do not tamper with NEWS.. You don't know what next release is going to hold, and nor do you know the date of the release. As long as commits are clear. If you assume this issue is going be fixed for 17.3.0, then update the Jira. Thx

adamdickmeiss (changes requested):
I consistently get unit test failures on this branch and not on master. Something is not right:
```
[ERROR] Errors: 
[ERROR]   UsersAPIIT.beforeAll:54->postTenant:111 » ConditionTimeout Condition with lamb...
[INFO] 
[ERROR] Tests run: 6, Failures: 0, Errors: 1, Skipped: 0
```

Looks like the new introduced construct `await().until(() -> tenantFinishedLoading(id) == true);` doesn't wait too much..

For me using ` await().atMost(Duration.TEN_SECONDS).until(() -> tenantFinishedLoading(id));`  fixes it.

NEWS.md
marcjohnson-kint: @adamdickmeiss 

I agree that the date of the release is unknown. I provided this feedback to @K-Felk previously and he corrected it. I think the release version is relatively known, given that it is defined at the previous release and only a compatibility breaking change will affect that (at which point, the news can be updated).

It was agreed by the Technical Leads [last year|https://wiki.folio.org/display/TLG/2020-06-24+Meeting+notes] that it is [acceptable to change|https://wiki.folio.org/display/DD/Pull+Request+Guidelines] the news during a pull request at the discretion of the lead maintainer.

I am ok with (rather I prefer) the news being changed during pull requests, as it reduces the burden on the person making the release and allows the author of the change to describe it (rather than the releasor having to guess). I believe this is common practice for the front end modules.

I will raise a JIRA issue and pull request to make this policy explicit for this module. Apologies for the inconvenience.

pom.xml
```diff
@@ -517,12 +521,12 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-    <raml-module-builder.version>31.1.5</raml-module-builder.version>
-    <vertx.version>3.9.4</vertx.version>
+    <raml-module-builder.version>32.0.0</raml-module-builder.version>
```
marcjohnson-kint, line 524: Might be worth going to `32.0.2` given that has now been released.

src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -1,48 +1,40 @@
...
-        .perform(ta, headers, vertx, res1 -> {
-          if (res1.failed()) {
-            hndlr.handle(io.vertx.core.Future.succeededFuture(PostTenantResponse
-              .respond500WithTextPlain(res1.cause().getLocalizedMessage())));
-            return;
-          }
-          hndlr.handle(io.vertx.core.Future.succeededFuture(PostTenantResponse
-            .respond201WithApplicationJson("")));
+  Future<Integer> loadData(TenantAttributes attributes, String tenantId,
+                           Map<String, String> headers, Context vertxContext) {
```
marcjohnson-kint, line 20: Please use the standard line continuation indentation. I would also leave a whitespace gap between a multiple line method signature and the implementation.

```suggestion
  Future<Integer> loadData(TenantAttributes attributes, String tenantId,
    Map<String, String> headers, Context vertxContext) {
    
```

src/test/java/org/folio/moduserstest/RestVerticleIT.java
```diff
@@ -89,61 +91,42 @@
...
-          });
-        } catch (Exception e) {
-          context.fail(e);
-        }
-      });
+        .setConfig(new JsonObject().put("http.port", port));
+
+    vertx.deployVerticle(RestVerticle.class.getName(), options, context.asyncAssertSuccess(res -> {
+      TenantAttributes ta = new TenantAttributes();
+      ta.setModuleTo("mod-users-1.0.0");
```
marcjohnson-kint, line 112: @julianladisch @adamdickmeiss Does the `toModule` property still need to be set when invoking the Tenant API?

I thought RAML Module Builder now tracked versions internally.

src/test/java/org/folio/rest/impl/UsersAPIIT.java
```diff
@@ -91,20 +103,35 @@ static TenantAttributes tenantAttributes() {
...
 
   static void postTenant() {
-    given().body(tenantAttributes()).
+    String id = given().body(tenantAttributes()).
     when().post("/_/tenant").
-    then().statusCode(201);
+    then().statusCode(201).
+    extract().
+    path("id");
+    await().until(() -> tenantFinishedLoading(id) == true);
```
marcjohnson-kint, line 111: Does await have a default timeout in case the tenant initialisation never succeeds?

src/test/java/org/folio/moduserstest/RestVerticleIT.java
adamdickmeiss: It should be there to indicate create (both from + to to indicate upgrade). RMB will not care about the values. But others might.

src/test/java/org/folio/rest/impl/UsersAPIIT.java
adamdickmeiss: I'm sure it has,. But why bother? await is the wrong solution to the problem.. It will fire HTTP GETs like mad.. We can just tell the async to return within 60 seconds or other.

src/test/java/org/folio/moduserstest/RestVerticleIT.java
marcjohnson-kint: Thanks @adamdickmeiss for clarifying those expectations.

@K-Felk please could a comment be added to explain that the version doesn't matter to RAML Module Builder, and this is used as a marker for a new activation rather than an upgrade.

pom.xml
adamdickmeiss: Or RMB 32.1.0 which was also released and has more features.

src/test/java/org/folio/rest/impl/UsersAPIIT.java
marcjohnson-kint: > We can just tell the async to return within 60 seconds or other.

Are you referring to the async within the RAML Module Builder implementation?

pom.xml
marcjohnson-kint: Even better, thanks @adamdickmeiss

src/test/java/org/folio/rest/impl/UsersAPIIT.java
adamdickmeiss: No.

https://github.com/folio-org/raml/blob/d9ab6f377b24530ffbe92129caf63f5cb0814e7a/ramls/tenant.raml#L48

marcjohnson-kint: Oh, the client can effectively make the (asynchronous by default) tenant API block till complete or a wait time has elapsed?

adamdickmeiss: yes. It as added so client did not have to poll too often (especially when we test).

marcjohnson-kint: @adamdickmeiss Ok, thanks. What wait time would you advise?

adamdickmeiss: 1 minute (60000) is fine. In some cases, it takes long time for Postgres to start.

marcjohnson-kint: Ok. The client should wait for at least that long as well then.

@K-Felk please change this code to use the waiting variant of the tenant API as @adamdickmeiss advises with a timeout of 60 seconds / 1 minute.

adamdickmeiss: Yes.. I have posted the exact code to do that earlier for inspiration.

marcjohnson-kint: Thanks @adamdickmeiss

pom.xml
K-Felk: POM is updated.  I have also removed the Awaitility import.

src/main/java/org/folio/rest/impl/TenantRefAPI.java
K-Felk: Done.

src/test/java/org/folio/moduserstest/RestVerticleIT.java
K-Felk: This is done, you may want to check the comment to make sure it reflects this discussion accurately.

src/test/java/org/folio/rest/impl/UsersAPIIT.java
K-Felk: Changed.  Tests seem to be passing on my end, but they did before when using Await, so...

marcjohnson-kint (approved):

adamdickmeiss (commented):
Much better. Now you just have a merge to consider. Let me know if you need help with that.

You have some trailing whitespace here and there.. Please try to avoid that - so that we don't have to see "red" when we do ``git diff`` :-)

K-Felk:
> Much better. Now you just have a merge to consider. Let me know if you need help with that.

I've fixed it locally by merging the 242 branch into master (after pulling to make sure master was up-to-date).  Re-ran tests, things look okay.  From here I'm assuming I can just push directly to master? 


> You have some trailing whitespace here and there.. Please try to avoid that - so that we don't have to see "red" when we do `git diff` :-)

I try to be careful about this, but for this task specifically, I was messing (and re-messing) with a lot of code, trying to figure out where and why things were going wrong.

adamdickmeiss:
> > Much better. Now you just have a merge to consider. Let me know if you need help with that.
> 
> I've fixed it locally by merging the 242 branch into master (after pulling to make sure master was up-to-date). Re-ran tests, things look okay. From here I'm assuming I can just push directly to master?

No. Stay in FOLIO-242 branch and merge master to it.

```
git fetch
git checkout FOLIO-242
git merge origin/master
# fix conflicts
git push

```

> 
> > You have some trailing whitespace here and there.. Please try to avoid that - so that we don't have to see "red" when we do `git diff` :-)
> 
> I try to be careful about this, but for this task specifically, I was messing (and re-messing) with a lot of code, trying to figure out where and why things were going wrong. Hopefully next time I won't need to monkey around as much.

adamdickmeiss (approved):
Thx for your patience

K-Felk:
Thanks for yours!

PR #204, opened 2021-01-30
Expire users with Okapi timer (1 minute) fixes MODUSERS-244

adamdickmeiss:
This has been tested on a Vagrant box .. where existing mod-users upgraded to this PR code. There's no behavioral change, except that I thought 1 minute would better than 30 seconds.. Can of course be adjusted.


ramls/users.raml
```diff
@@ -51,3 +51,17 @@ resourceTypes:
...
+  /expire/timer:
+    post:
+      description: Expire timer (timer event)
+      responses:
+        204:
+          description: Expired ok
+        400:
+          description: User Error
+          body:
+            text/plain:
```
hjiebsco, line 63: Is this needed?  There is no implementation to return 400.

hjiebsco (approved):
Nice change! Glad to see `select nspname from pg_catalog.pg_namespace where nspname ~ '^[^_]+_mod_users$';` is gone.

ramls/users.raml
adamdickmeiss: RMB may return it - before our handler is called.

adamdickmeiss:
> Nice change! Glad to see `select nspname from pg_catalog.pg_namespace where nspname ~ '^[^_]+_mod_users$';` is gone.

Yes... This code is old, added May 31 2018 , so it first appeared in release mod-users 15.2.0.

Okapi started to offer timer in May 2019.

PR #207, opened 2021-05-26
MODUSERS-122: DELETE /users with CQL

julianladisch:


ramls/users.raml
```diff
@@ -44,6 +44,25 @@ resourceTypes:
   post:
     is: [validate]
     description: Create a user
+  delete:
+    description: Delete a collection of users selected by a CQL query; this doesn't delete proxyFor records that reference them
+    is: [
+      language,
```
marcjohnson-kint, line 50: Given that FOLIO is close to deciding to use a header for language localization, why introduce a query string parameter at this point that will likely need to be removed soon anyway?

julianladisch: Removed. (This pull request is more than 3 months old.)

marcjohnson-kint: Thank you for removing it

src/main/java/org/folio/rest/impl/UsersAPI.java
```diff
@@ -299,6 +298,15 @@ public void deleteUsersByUserId(String userId, String lang,
       DeleteUsersByUserIdResponse.class, asyncResultHandler);
   }
 
+  @Validate
+  @Override
+  public void deleteUsers(String query, RoutingContext routingContext,
+          Map<String, String> okapiHeaders,
+          Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
+    PgUtil.delete(getTableName(null), query, okapiHeaders, vertxContext,
```
marcjohnson-kint, line 306: Pesonally, I would leave a gap between a multi-line method signature and the implementation (and reduce the indent, which only takes up horizontal space)

```suggestion
    Map<String, String> okapiHeaders,
    Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
    
    PgUtil.delete(getTableName(null), query, okapiHeaders, vertxContext,
```

src/test/java/org/folio/rest/impl/UsersAPIIT.java
```diff
@@ -121,8 +121,28 @@ static void postTenant() {
...
+  }
+
+  static void postUser(String id, String username) {
+    given().
+    when().
+      body(new JsonObject().put("id",  id).put("username", username).encode()).
+      post("/users").
+    then().
+      statusCode(201);
+  }
```
marcjohnson-kint, line 143: Why do these need to be static?

Personally, I think the tests are the star of the show and any helpers (which I assume these are) are intended to make the tests more readable are better kept out of the way at the bottom of the file.

src/test/java/org/folio/rest/impl/UsersAPIIT.java
```diff
@@ -145,4 +165,23 @@ void facetsLimit0() {
   void facetsLimit1() {
     facets(1);
   }
+
+  @Test
+  void delete() {
```
marcjohnson-kint, line 170: Personally, I think it would be better (less ambiguous, more intent revealing, better describing) to be more specific that this test is to delete multiple users (using CQL) rather than a single user.

```suggestion
  void deleteMultipleUsersUsingCQL() {
```
```suggestion
  void multipleUsersCanBeDeletedUsingCQL() {
```

src/test/java/org/folio/rest/impl/UsersAPIIT.java
```diff
@@ -145,4 +165,23 @@ void facetsLimit0() {
...
+    String id1 = "114b14e8-422d-429a-b89e-b55c439b3b21";
+    String id2 = "22531038-08b8-4c71-a72c-9d48f63f3682";
+    String id3 = "13836e0e-4331-4386-a8be-2345eece6de3";
```
marcjohnson-kint, line 173: Personally, I wouldn't hardcode these and would use a randon UUID each time.

src/test/java/org/folio/rest/impl/UsersAPIIT.java
```diff
@@ -121,8 +121,28 @@ static void postTenant() {
...
+    then().statusCode(200);
+  }
+
+  static void userDoesntExist(String id) {
+    given().
+    when().get("/users/" + id).
+    then().statusCode(404);
+  }
+
+  static void postUser(String id, String username) {
```
marcjohnson-kint, line 136: I believe the intent is to create a user. I think it is a detail that this helper is deliberately trying to hide that it is done via a POST request.

```suggestion
  static void createUser(String id, String username) {
```

src/test/java/org/folio/rest/impl/UsersAPIIT.java
```diff
@@ -145,4 +165,23 @@ void facetsLimit0() {
...
+    String id3 = "13836e0e-4331-4386-a8be-2345eece6de3";
+    postUser(id1, id1);
+    postUser(id2, id2);
+    postUser(id3, id3);
+    given().
+    when().
+      param("query", "username == \"1*\"").
+      delete("/users").
+    then().
+      statusCode(204);
```
marcjohnson-kint, line 182: Given that helpers have been extracted for all of the other steps. Could one be extracted for this too?

I think it would help the test to read at the same level of abstraction.

```suggestion
    deleteUsersByUsername("1*")
```

src/test/java/org/folio/rest/impl/UsersAPIIT.java
```diff
@@ -145,4 +165,23 @@ void facetsLimit0() {
...
+    userExists(id2);
+    userDoesntExist(id1);
+    userDoesntExist(id3);
```
marcjohnson-kint, line 185: Personally, I'd use some whitespace for differentiate between the arrangement, the act and the checks.

```suggestion
    String id1 = "114b14e8-422d-429a-b89e-b55c439b3b21";
    String id2 = "22531038-08b8-4c71-a72c-9d48f63f3682";
    String id3 = "13836e0e-4331-4386-a8be-2345eece6de3";
    
    postUser(id1, id1);
    postUser(id2, id2);
    postUser(id3, id3);
    
    given().
    when().
      param("query", "username == \"1*\"").
      delete("/users").
    then().
      statusCode(204);
      
    userExists(id2);
    userDoesntExist(id1);
    userDoesntExist(id3);
```

src/main/java/org/folio/rest/impl/UsersAPI.java
julianladisch: The new code mimics the other methods in this file for consistency.

Changing the formatting of the whole file is out of scope of this pull request, please create an issue against mod-users for this code cleanup task.

src/test/java/org/folio/rest/impl/UsersAPIIT.java
julianladisch: This is an integration test that also tests the POST /user API. This helper doesn't try to hide anything but to avoid duplicate code.

src/test/java/org/folio/rest/impl/UsersAPIIT.java
julianladisch: The static keyword indicates that the method doesn't use any instance variables, and it can be used in the static `@BeforeAll` and `@AfterAll` methods if at some time a need arises.

julianladisch: The new helper methods are put at the same place where the existing helper method facets(int) already exists for consistency.

If you want the order been changed from "helpers - tests" to "tests - helpers" please create an issue so that it can be handled separately.

src/test/java/org/folio/rest/impl/UsersAPIIT.java
julianladisch: username == "1*" only works as intended if id1 and id3 start with 1 and id2 don't.

Using a random UUID doesn't result in consistent tests, sometimes it fails, sometimes it succeeds.

src/test/java/org/folio/rest/impl/UsersAPIIT.java
julianladisch: Empty line inserted before the tests.

src/test/java/org/folio/rest/impl/UsersAPIIT.java
marcjohnson-kint: > This is an integration test that also tests the POST /user API. This helper doesn't try to hide anything but to avoid duplicate code.

I don't understand why the type of tests changes the value of expressing intent to improve the readability of the code.

Are you telling me that you are rejecting my suggestion as irrelevant?

src/test/java/org/folio/rest/impl/UsersAPIIT.java
marcjohnson-kint: I suggest defining the details that the tests relies on explicitly, in this case the username instead of inferring it from something the record ID that is incidental and should be different each time (as this is what we would expect in real use).

src/test/java/org/folio/rest/impl/UsersAPIIT.java
marcjohnson-kint: > The static keyword indicates that the method doesn't use any instance variables, and it can be used in the static @BeforeAll and @AfterAll methods if at some time a need arises.

Why preemptively mark them as static if it's not needed now?

> If you want the order been changed from "helpers - tests" to "tests - helpers" please create an issue so that it can be handled separately.

I think that is an unnecessary overhead and is unlikely to lead to these changes being made.

src/main/java/org/folio/rest/impl/UsersAPI.java
marcjohnson-kint: I didn't suggest reformatting the entire file. I suggested making one simple change to improve the readability of code that was already being changed in this pull request.

julianladisch: The pull request uses 8 space indentation to comply with https://www.oracle.com/java/technologies/javase/codeconventions-indentation.html#248 suggested by https://dev.folio.org/guidelines/contributing/#style-conventions . 8 space indentation is consistently used throughout UsersAPI.java.

You suggest to switch to 4 space indentation suggested by https://google.github.io/styleguide/javaguide.html#s4.5.2-line-wrapping-indent . This code style change should be made throughout UsersAPI.java for consistency, not only for a single method.

src/test/java/org/folio/rest/impl/UsersAPIIT.java
julianladisch: The method name `postUser` indicates that the POST /user API is used to create the user.
The method name `createUser` doesn't and the `createUser` method may use any implementation to create the user, for example an SQL query.
`postUser` might be used for tests we may add in the future where the POST /user API is the API under test.
Therefore I want the method name to be `postUser` and not `createUser`.

src/main/java/org/folio/rest/impl/UsersAPI.java
marcjohnson-kint: > 8 space indentation is consistently used throughout UsersAPI.java.

That is unfortunate, as [FOLIO decided](https://dev.folio.org/guidelines/contributing/) (quoted below) long ago to diverge from the Oracle standard and use 2 spaces rather than 4 per indentation.

>  _We indent with two spaces only, because vert.x uses deeply nested callbacks._

marcjohnson-kint: I care more about the vertical whitespace change. Do you also object to that?

src/test/java/org/folio/rest/impl/UsersAPIIT.java
marcjohnson-kint: > Why should I declare the new helper methods to be non-static when they in fact are also static? Isn't this inconsistent and misleading?

Why would it having a method that doesn't need to be static, be non-static be misleading?

Personally, I think methods should be declared static when they need to be. If they aren't used in a static context, I don't think they need to be static.

I would also put static helper methods below the tests (which for me are the purpose of the class).

However, I understand that for most folks static methods belong at the top of the class (and that overrides the prominence of the tests for them) and they prefer to put methods close to where they are used.

The three examples you cite are for the purposes of shared setup. They need to be static because they are used in a static context.  I don't think all helper methods must be static by definition, nor is there is value in consistency with those only for the sake of it.

src/test/java/org/folio/rest/impl/UsersAPIIT.java
marcjohnson-kint: > The method name postUser indicates that the POST /user API is used to create the user.

Why, within this test, is the specifics of the mechanism for user creation important? 

To me, the intention is to create a user, not to exercise a particular endpoint. The goal of this part of the state is to arrange the required state for the rest of the test. How that is done is less relevant than what is trying to be achieved.

> postUser might be used for tests we may add in the future where the POST /user API is the API under test.

Hopefully there are already tests for that endpoint already. I don't think speculation about future use is a good argument for why it is more important to be specific about the mechanism rather than the intention.

> Therefore I want the method name to be postUser and not createUser.

I understand that you disagree with my feedback. It is your change and so it is your preference. You can reject the feedback.

src/main/java/org/folio/rest/impl/UsersAPI.java
julianladisch: Yes, because the body is a single statement that is a method call. This is easy to grasp, in such cases I never add a blank line:

```
  public void deleteUsers(String query, RoutingContext routingContext,
      Map<String, String> okapiHeaders,
      Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
    PgUtil.delete(getTableName(null), query, okapiHeaders, vertxContext,
         DeleteUsersResponse.class, asyncResultHandler);
  }
```

I don't see an improvement when inserting the blank line:

```
  public void deleteUsers(String query, RoutingContext routingContext,
      Map<String, String> okapiHeaders,
      Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {

    PgUtil.delete(getTableName(null), query, okapiHeaders, vertxContext,
         DeleteUsersResponse.class, asyncResultHandler);
  }
```

However, if the body consists of multiple statements or a construct like a loop or if-then-else then I add a blank line because it is more likely that the line-wrapped method parameters get misread as a wrapped statement.

marcjohnson-kint: > Yes, because the body is a single statement that is a method call. This is easy to grasp, in such cases I never add a blank line:

I find it difficult to differentiate between the multi-line method signature (which has indentation) and the body of the method. To me, this also applies when the body is only a single statement (especially when that statement is split across multiple lines).

I'll admit, your answer surprises me, as I think I recall a conversation we had where you asked for whitespace between (dependency maybe) entries in a pom file in order to help distinguish between them.

Anyway, I'll take this comment as another rejection of my feedback.

zburke:
@julianladisch, @marcjohnson-kint, you both make some really well-argued points about formatting on this PR that, nonetheless, are entirely beside the point of the PR itself. On the front-end, we've resolved this kind of debate by strongly enforcing a set of [shared ESLint rules](https://github.com/folio-org/eslint-config-stripes). This has had several positive effects: 

* Conversations about style happen on PRs in the repository where rules are defined, which feels like the right place for them. It also gives the entire community the opportunity to review and comment on changes before they become law.
* It avoids hijacking PRs (like this one) with conversations about style.
* It prevents unresolvable discussions about personal preferences. 
* Consistent formatting across repos makes code more approachable to newcomers and reduces the cognitive load for people who already manage multiple projects. 

The [contributing guidelines](https://dev.folio.org/guidelines/contributing/) Marc mentioned are insufficiently specific and don't appear to be enforceable because aren't part of a config file for Checkstyle/PMD/Coala/whatever. I know this has come up a few times in the past at tech leads and it always gets shot down. Frankly, I view that as a tremendous missed opportunity, not because I love forcing people to format code in line with my personal preferences, but because we have some big, thorny problems on this project that I think are more deserving of our attention and talent than how we manage whitespace. I'll bring it up at tech leads again in a few weeks unless somebody else decides to write a decision log entry or write a commit to checkstyle-config-folio or something else.

adamdickmeiss (approved):

marcjohnson-kint:
> @julianladisch, @marcjohnson-kint,

My TLDR on this topic.

The community deferred this decision for a long time. Some developers went their own way (some aligned some actively contradictory to the guidance). 

Then we almost decided to adopt Google's code style (which is fairly flexible so wouldn't resolve all of these conversations). However the amount of work required to retrofit that to existing modules so it's unlikely we'll decide it officially and enact it soon.

zburke:
@marcjohnson-kint, thanks for the summary/history. It's helpful to have that context. Nonetheless, I think there should be a "from this point forward" decision as well as a more strict set of rules if Google's are so lax they "wouldn't resolve all of these conversations." Code standards are less about picking The One True Style (though I'm sure some folks think they are) and more about establishing and sticking to community standards so we don't have to spend time re-litigating them on every encounter. I prefer driving on the left, personally, but for the good of my community I don't argue this point with every passing motorist. 

Many of us work with universities. Can we find some interns who want experience contributing to a major open source project and hand them the reformatting job?

adamdickmeiss:
> @marcjohnson-kint, thanks for the summary/history. It's helpful to have that context. Nonetheless, I think there should be a "from this point forward" decision as well as a more strict set of rules if Google's are so lax they "wouldn't resolve all of these 

I've introduced Google style on projects and it solves, perhaps, 80% of the issues. At least I don't find myself discussing code style much.. Which is a waste of time.

marcjohnson-kint:
> I've introduced Google style on projects and it solves, perhaps, 80% of the issues. At least I don't find myself discussing code style much.. Which is a waste of time.

I'm in agreement with that, especially for new modules. (it does need to be well documented so developers can easily run it in a local build, I recently had a pull request fail because I didn't know a module enforced Google style)

This does currently actively contravene FOLIO's official guidance. I'd welcome FOLIO officially supporting Google's Java style and using checkstyle to validate it.

marcjohnson-kint:
> Many of us work with universities. Can we find some interns who want experience contributing to a major open source project and hand them the reformatting job?

I like this idea. My primary concerns with these efforts is the conflict they tend to have with ongoing work and the time it takes to review them.

PR #214, opened 2021-08-13
MODUSERS-247: Add unique constraint for user externalSystemId

julianladisch:
The Check-Out settings allow to use the external system id as a unique patron identifier: https://folio-snapshot.dev.folio.org/settings/circulation/checkout

If external system id is not unique the check-out will pick one of the multiple matches without any warning that it might be the wrong patron.

The upgrade will fail if existing patrons have non-unique external system id with an error message containing the duplicate value.

There is no automated way to fix duplicates during migration therefore the failure is the only option we have.


craigmcnally (approved):
looks pretty straightforward to me...

craigmcnally:
Question - do we also need migration scripts for this?  is it necessary to actively look for duplicates and if present fail the upgrade?  Or will postgres do this for us when adding the unique index?  In any case, duplicates need to be removed prior to upgrade.

adamdickmeiss (commented):
What a stupid constraint. My 5c

julianladisch:
> Question - do we also need migration scripts for this? is it necessary to actively look for duplicates and if present fail the upgrade? Or will postgres do this for us when adding the unique index? In any case, duplicates need to be removed prior to upgrade.

Does the discussion on https://issues.folio.org/browse/MODUSERS-247 and the [Kiwi Release Notes in the section Important Upgrade Considerations – Changes and Required Actions](https://wiki.folio.org/display/REL/R3+2021+%28Kiwi%29+Release+Notes#R32021(Kiwi)ReleaseNotes-ChangesandRequiredActions) answer your questions?

marcjohnson-kint:
@julianladisch This pull request is more than 6 months old. What is the intention with keeping this pull request open?

PR #215, opened 2021-08-20
MODUSERS-260, MODUSERS-261: Remove jsoup HTML stripping

julianladisch:
Remove jsoup HTML stripping.

It is not needed because Stripes/React can handle
special characters and prevent HTML injection.

Stripping characters must be declared in the RAML schema
https://github.com/folio-org/mod-users/blob/master/ramls/department.json
after the User Management SIG and the PO have agreed.

If some characters should be removed or replaced
the backend must do it consistently. This is not the case,
see examples on https://issues.folio.org/browse/MODUSERS-260

The used version of jsoup has a security vulnerability:
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-37714


marcjohnson-kint:
It seems the stripping of department properties was deliberate. Do we know why this was introduced?

I think we need to understand that before it is decided that it be removed.

psmagin (approved):

julianladisch:
@marcjohnson-kint : Please explain in more detail why you don't understand that the front-end, not the back-end, is the right place to prevent HTML injection. Pavel Smagin (psmagin), the author of the stripping code, has already agreed that it should be removed from mod-users.

marcjohnson-kint:
@julianladisch 

> Please explain in more detail why you don't understand that the front-end, not the back-end, is the right place to prevent HTML injection. 

I have made no such assertion.

I only asked about the history of why this code was included originally, as I think it is important to understand why something was done in order to decide if it should be undone. 

How did my question come across as an assertion?

> Pavel Smagin (psmagin), the author of the stripping code, has already agreed that it should be removed from mod-users.

If you are referring to his approval of this pull request, that happened 2 days after I asked my question, so I couldn't have known about it at the time.

If you don't think my question is relevant (either ever or with that newer information), please disregard it.

Either way, I think it makes sense for me to leave this conversation. I apologise for getting involved :-/

zburke:
A+ for @marcjohnson-kint here on the topic of understanding why something is present as a prerequisite to removing it. This is [Chesterton's Fence](https://fs.blog/2020/03/chestertons-fence/), and good fences make good neighbors.

The examples of mangled input in [MODUSERS-260](https://issues.folio.org/browse/MODUSERS-260) are a compelling example that there is a problem to be fixed, but the original ticket, [MODUSERS-158](https://issues.folio.org/browse/MODUSERS-158) specifically mentions preventing "code/scripting in field". Indeed, as @julianladisch mentions, React will escape this kind data, but a React app may not be the only client of this API. Do we know why this was specifically mentioned there, given it seems to be an unusual requirement?

If "the backend must do [escaping] consistently", is there a story for that? Are there PRs for that? If so, it would be great to provide links to them to provide context for this work. If not, isn't it premature to remove this code? I found my way here via [MODINV-475](https://issues.folio.org/browse/MODINV-475), which the support SIG discussed earlier this week. On that ticket, various POs all agreed that trimming whitespace, if not all "potentially dangerous" characters, should be a general function of the backend. Front-end developers at #stripes-architecture agreed. What do you think is the right forum to start a conversation input processing and how the backend could do it consistently?

PR #217, opened 2021-09-10
MODUSERS-262: Validate patronGroup and addressTypeId to be UUID

julianladisch:
Fix HTTP status code and return a body with good description what
is wrong.


adamdickmeiss (approved):

marcjohnson-kint:
Given that this restricts the values allowed by those properties meaning previously allowed values are not allowed any more.

Why was this not considered a breaking compatibility change for the `users` interface?

PR #219, opened 2021-09-15
MODUSERS-265: Reformat methods with multi-line parameter list

julianladisch:
If the parameter list of a method is broken into multiple lines indent by 4 spaces and put an empty line after the parameter list.

Requested on https://github.com/folio-org/mod-users/pull/207

This complies with

* https://google.github.io/styleguide/javaguide.html#s4.5.2-line-wrapping-indent and
* https://www.oracle.com/java/technologies/javase/codeconventions-indentation.html#248 and
* https://dev.folio.org/guidelines/contributing/#consistent-whitespace


steveellis (approved):

adamdickmeiss (approved):
I'm sure I have created many of these. The default IntelliJ IDEA settings with editorconfig doesn't do it. But I found the correct setting now . Thx

PR #223, opened 2021-10-04
MODUSERS-272: Update RMB to 33.1.1 and Vert.x to 4.1.4

julianladisch:
RMB 33.1.1 is needed to fix the database memory fix RMB-863. mod-users is affected because it uses PgUtil.streamGet: https://github.com/folio-org/mod-users/blob/v18.1.0/src/main/java/org/folio/rest/impl/UsersAPI.java#L149

Also update folio-service-tools.version to 1.7.1 and
folio-custom-fields.version to 1.6.1
for Vert.x 4.1.4 support.

Also update joda-time from 2.10.6 to 2.10.12 and some test dependencies.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-## 18.1.1 2021-10-04
+## 18.1.1
 
 * Update RMB to 33.1.1 and Vert.x to 4.1.4 (MODUSERS-272)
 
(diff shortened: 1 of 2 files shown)


julianladisch:
This is not the release.
For the release we have https://issues.folio.org/browse/MODUSERS-271 "Release v18.1.1 for Kiwi R3 2021", and I will release mod-users 18.1.1 today.
I've removed the release date from NEWS.md to indicate that this PR doesn't release mod-users.

adamdickmeiss (approved):

PR #233, opened 2022-01-25
Modusers 301

rladdusaw:
Resolves [MODUSERS-301](https://issues.folio.org/browse/MODUSERS-301)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 ## 18.2.0 IN-PROGRESS
 
 * Upgrade to RMB 33.1.3 and Folio Service Tools 1.7.2. (CVE-2021-44228) (MODUSERS-295)
+* Upgrade to RMB 33.2.4 (MODUSERS-301)
 
 ## 18.1.1 2021-10-04
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,6 @@
 ## 18.2.0 IN-PROGRESS
 
-* Upgrade to RMB 33.1.3 and Folio Service Tools 1.7.2. (CVE-2021-44228) (MODUSERS-295)
-* Upgrade to RMB 33.2.4 (MODUSERS-301)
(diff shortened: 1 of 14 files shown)


NEWS.md
```diff
@@ -1,6 +1,7 @@
 ## 18.2.0 IN-PROGRESS
 
 * Upgrade to RMB 33.1.3 and Folio Service Tools 1.7.2. (CVE-2021-44228) (MODUSERS-295)
+* Upgrade to RMB 33.2.4 (MODUSERS-301)
```
marcjohnson-kint, line 4: I would combine these two lines. I don't think it is necessary to describe RMB upgrades twice in the release news

```suggestion
* Upgrade to RMB 33.2.4 and Folio Service Tools 1.7.2. (CVE-2021-44228) (MODUSERS-295, MODUSERS-301)
```

K-Felk (commented):
I'm puzzled here-I see you doing things like updating the ramls, removing the lang parameter from function calls, etc., which is what I would expect.  But then you revert all that, and just change the pomfile.  Can you tell me why?  I'm struggling a bit with this work in modinvstor, so I really would like to understand.

K-Felk:
> Does that make sense?

ah!  so not only do you not have to change the RAMLs, you also *don't have to remove the lang parameter from function calls* either.  I think it was this line in the instructions that was throwing me:

>Therefore you need to change the parameters of your methods that implement the RAML generated interfaces - remove >the lang parameter, and add the String totalRecords parameter before the int offset parameter. The compile will fail >unless this is done

If you remove the lang param from function calls, then you have to remove it from the RAML files.  Which in turn means you have to do something about references to the parameter in the raml-util subrepository, which is what I was struggling with.

marcjohnson-kint (approved):

marcjohnson-kint:
> > Does that make sense?
> 
> ah! so not only do you not have to change the RAMLs, you also _don't have to remove the lang parameter from function calls_ either.

Yeah, effectively all of the steps in the upgrade notes are coupled together. Either the submodule, RAMLs and code need changing or none do, which I personally didn't find obvious. And there is a note at the bottom, that I totally missed, that states it is optional.

PR #240, opened 2022-03-31
MODUSERS - 306 POST and PUT of user object fail with error "Invalid Text format" if the object contains a non-required customFields of type text and the value is an empty string

DieHardJava:
### Purpose 
POST and PUT of user object fail with error "Invalid Text format" if the object contains a non-required customFields of type text and the value is an empty string

### Approach
- added method removeCustomFieldIfEmpty which removes customfield if object value is empty before validation in postUsers and putUsersById methods
- wrote unit test for removeCustomFieldIfEmpty to check


src/main/java/org/folio/rest/impl/UsersAPI.java
```diff
@@ -157,6 +157,13 @@ public void getUsers(String query, String orderBy,
     }
   }
 
+  void removeCustomFieldIfEmpty(User entity) {
```
psmagin, line 160: Make it private

src/main/java/org/folio/rest/impl/UsersAPI.java
```diff
@@ -157,6 +157,13 @@ public void getUsers(String query, String orderBy,
...
+    var customField = (entity.getCustomFields() != null) ?
+      entity.getCustomFields().getAdditionalProperties() : null;
+    if (customField != null)
+      customField.entrySet().removeIf(obj -> obj.getValue().toString().isEmpty());
```
psmagin, line 164: add check for `value instaceof String`

src/test/java/org/folio/rest/impl/UsersAPITest.java
```diff
@@ -91,6 +93,15 @@ void postUsersException(VertxTestContext vtc) {
...
 
+  @Test
+  void removeCustomField(VertxTestContext vtc) {
+    CustomFields customFields = new CustomFields().withAdditionalProperty("test", "");
+    User user = new User().withCustomFields(customFields);
+    new UsersAPI().removeCustomFieldIfEmpty(user);
+    vtc.verify(() -> assertTrue(user.getCustomFields().getAdditionalProperties().isEmpty()));
+    vtc.completeNow();
+  }
+
```
psmagin, line 104: It should be an API test. Please, check how API tests are written in other test cases.

psmagin (changes requested):

psmagin (approved):

Igor-Gorchakov (approved):

Shans-Kaluhin (approved):

PR #241, opened 2022-04-22
MODUSERS-254 Work in support of Sip2-89 

ianibo:
Provide for separate secure storage of hash derived from a user input pin. 
Provide an endpoint to validate some arbitrary pin against the derived hash and return an appropriate status allowing client apps (E.G. Edge-Sip) to determine if the user is authentic.

Please feel free to squash merge and  provide sane history.


sosguthorpe (approved):
LGTM

PR #242, opened 2022-05-12
Remove user facets API MODUSERS-308

marcjohnson-kint:
## Purpose

[UIU-1562](https://issues.folio.org/browse/UIU-1562) removed the need for a count of users by patron group from the UI (as the current implementation was not fit for purpose).

In order to reduce the maintenance burden, the facets functionality can be removed from mod-users.

## Approach
* Remove the user facets API
* Increment major version of `users` interface (as this is an breaking change)
* Increment module major version (as this is an breaking change)
* Stop loading sample records during users API integration test (as tests no longer rely on them)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 ## 19.0.0 IN-PROGRESS
 
 * Made AddressTypeId field required for all user addresses (MODUSERS-212)
+* Removes users facets API (MODUSERS-308)
 * Provides `users 16.0`
 
 ## 18.3.0 2022-06-13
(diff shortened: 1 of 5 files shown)


K-Felk (approved):

descriptors/ModuleDescriptor-template.json
```diff
@@ -4,7 +4,7 @@
   "provides": [
     {
       "id": "users",
-      "version": "15.5",
+      "version": "16.0",
```
julianladisch, line 7: Do we know if ui-users is the only module that uses this API's facets?

marcjohnson-kint: For the centralised modules, I believe so.

I did a search for facets in github and checked all of the code that referenced the term. The only one I found that refers to them users facets is the code that has now been removed from ui-users

julianladisch (approved):
Looks good, thanks!

PR #245, opened 2022-05-19
Improve users integration tests MODUSERS-309

marcjohnson-kint:
*Purpose*

Reduce the number of [reported code smells](https://sonarcloud.io/project/issues?resolved=false&id=org.folio%3Amod-users)

*Approach*
* Extract common behaviour for deploying module and enabling for a tenant to a separate class
* Stop the current users tests relying on sample data
* Move the tests relating to users from the groups to users integration tests
* Move reference records to within the code (to avoid a maven build step to copy them)
* Introduce a Patron Pin client (much like for users and clients) to abstract API requests
* Split up the existing test in individual tests (to reduce complexity and improve feedback)
* Split up patron pin related tests from general users tests
* Resolve unexpected failure of patron pin API when pin record does not exist (fetching the record returns success with a null record rather than a failure)

*TODO*
*  (requires improvement to that test class first)
* Reduce complexity of rest verticle integration tests by splitting examples into separate tests


adamdickmeiss (commented):
Good to move reference-data to resources. They started being here for historical reasons.. import.sh scripts.. Also in mod-inventory-storage. mod-users is git submodule-included by other modules, but as far as I can tell only sample-data is used - and that's in resources already.

src/test/java/org/folio/rest/impl/UsersAPIIT.java
```diff
@@ -1,99 +1,120 @@
...
+    groupsClient = new GroupsClient(new URI("http://localhost:" + port), headers);
+
+    final var module = new VertxModule(vertx);
 
-  @AfterAll
-  static void afterAll() {
-    vertx.close();
+    module.deployModule(port)
+      .onComplete(context.succeeding(res -> module.enableModule(headers,
+          true, false)
```
adamdickmeiss, line 68: This is the only place where enableModule loads reference data. Perhaps another test or this one to also load sample data and check that *something* is loaded.

marcjohnson-kint: Agreed. I'll add a test specifically for loading the sample data.

The existing tests only really check the sample data very superficially or accidentally.

marcjohnson-kint:
> They started being here for historical reasons.. import.sh scripts.. Also in mod-inventory-storage.

Yeah, the original version of those scripts were written when the modules didn't load sample data themselves. They could (likely should) have been removed when the modules started to load the data themselves.

wwelling (approved):

PR #246, opened 2022-05-25
Replace rest verticle integration tests MODUSERS-309

marcjohnson-kint:
*Purpose*

Reduce the number of [reported code smells](https://sonarcloud.io/project/issues?resolved=false&id=org.folio%3Amod-users)

*Approach*
* Move user related tests to users integration tests
* Separate out address type related tests to separate integration test class
* Separate out proxy relationship tests to separate integration test class
* Separate out user expiration tests to separate integration test class
* Remove unused code in Rest Support class


src/test/java/org/folio/moduserstest/ExpirationIT.java
```diff
@@ -0,0 +1,72 @@
...
+    usersClient = new UsersClient(new URI("http://localhost:" + port), headers);
+    expirationClient = new ExpirationClient(
+      new URI("http://localhost:" + port), headers);
+
+    final var module = new VertxModule(vertx);
+
+    module.deployModule(port)
+      .onComplete(context.succeeding(res -> module.enableModule(headers,
+          false, false)
+        .onComplete(context.succeedingThenComplete())));
```
julianladisch, line 53: module.deployModule(port)
      .compose(res -> module.enableModule(headers, false, false))
      .onComplete(context.succeedingThenComplete());

Using `.compose` and checking for failure at the end only helps me understanding the code because it is more idiomatic for Vert.x Futures.

src/test/java/org/folio/moduserstest/ProxyRelationshipsIT.java
```diff
@@ -0,0 +1,278 @@
...
+
+    final var relationship = proxiesClient.createProxyRelationship(
+      ProxyRelationship.builder()
+        .userId(userId)
+        .proxyUserId(proxyUserId)
+        .build());
+
+    assertThat(relationship.getId(), is(notNullValue()));
+    assertThat(relationship.getUserId(), is(userId));
+    assertThat(relationship.getProxyUserId(), is(proxyUserId));
```
julianladisch, line 84: Matching the rest-assured way:
```
    proxiesClient.createProxyRelationship(
       ProxyRelationship.builder()
         .userId(userId)
         .proxyUserId(proxyUserId)
        .build(),
      "id", is(notNullValue()),
      "userId", is(userId),
      "proxyUserId", is(proxyUserId));
```

It requires this `.body` call:
```
  public ProxyRelationship createProxyRelationship(
      @NonNull ProxyRelationship proxyRelationship,
      String path, Matcher<?> matcher, Object... pathMatcherPairs) {

     return attemptToCreateProxyRelationship(proxyRelationship)
        .statusCode(HTTP_CREATED)
        .body(path, matcher, pathMatcherPairs)
        .extract().as(ProxyRelationship.class);
  }
```

src/test/java/org/folio/moduserstest/ProxyRelationshipsIT.java
```diff
@@ -0,0 +1,278 @@
...
+        .userId(userId)
+        .proxyUserId(firstProxyUserId)
+        .build());
+
+    proxiesClient.attemptToCreateProxyRelationship(
+      ProxyRelationship.builder()
+        .userId(userId)
+        .proxyUserId(secondProxyUserId)
+        .build())
+      .statusCode(is(HTTP_CREATED));
```
julianladisch, line 123: Either add more tests or use the `.createProxyRelationship` that has built-in statusCode test.

src/test/java/org/folio/moduserstest/ProxyRelationshipsIT.java
```diff
@@ -0,0 +1,278 @@
...
+    final var errors = proxiesClient.attemptToCreateProxyRelationship(
+        ProxyRelationship.builder()
+          .userId(userId)
+          .proxyUserId(proxyUserId)
+          .build())
+      .statusCode(is(422))
+      .extract().as(ValidationErrors.class);
+
+    assertThat(errors.getErrors().get(0).getMessage(),
+      is("Proxy relationship already exists"));
```
julianladisch, line 166: A more rest-assured-like check:

```
      .statusCode(is(422))
      .body("errors[0].message", is("Proxy relationship already exists"));
```

src/test/java/org/folio/moduserstest/ProxyRelationshipsIT.java
```diff
@@ -0,0 +1,278 @@
...
+      String.format("userId==%s", userId));
+
+    assertThat(proxyRelationships.getTotalRecords(), is(2));
+
+    final var allProxies = proxyRelationships.getProxiesFor()
+      .stream()
+      .map(ProxyRelationship::getProxyUserId)
+      .collect(Collectors.toList());
+
+    assertThat(allProxies, containsInAnyOrder(firstProxyUserId, secondProxyUserId));
```
julianladisch, line 244: A more rest-assured-like test:
```
    proxiesClient.getProxyRelationships(String.format("userId==%s", userId),
        "totalRecords", is(2),
        "proxiesFor.collect { p -> p.proxyUserId }", containsInAnyOrder(firstProxyUserId, secondProxyUserId));
```
Requires this method:
```
  public ProxyRelationships getProxyRelationships(String cqlQuery,
      String path, Matcher<?> matcher, Object... pathMatcherPairs) {
    return given()
      .config(config)
      .spec(requestSpecification)
      .when()
      .queryParam("query", cqlQuery)
      .get("/proxiesfor")
      .then()
      .statusCode(HTTP_OK)
      .body(path, matcher, pathMatcherPairs)
      .extract().as(ProxyRelationships.class);
  }
```

src/test/java/org/folio/moduserstest/ExpirationIT.java
marcjohnson-kint: Thanks for the feedback. I'll try using compose instead.

I had trouble when I originally converted this code from the older vert.x integration and it was unstable and so kept much of the earlier structure of nesting callbacks.

src/test/java/org/folio/moduserstest/ProxyRelationshipsIT.java
marcjohnson-kint: I agree that RestAssured provides those capabilities. I disagree that it is _the_ rest assured way.

What is better about passing matchers into the client than expressing them separately?

I personally find it easier to understand the important aspects of the tests much of the mechanics of the API are abstracted away yet the pertinent checks are in the tests themselves.

I find it easier to read separate assertions and they tend to provide better feedback on failure, in that I know which specific assertion has failed.

There is a balance to be found in the use of RestAssured. A part of me would prefer to deserialise to a response as quickly as possible and do everything else separately. I decided that using RestAssured's status checking and body deseralization was a reasonable compromise.

This model also allows for model in the tests to adapt separately to the raw API's (partly to hide some unfortunate design decisions) and it supports refactoring tools better than using strings to refer to fields.

src/test/java/org/folio/moduserstest/ProxyRelationshipsIT.java
marcjohnson-kint: What additional tests would you want here?

I chose to attempt and an explicit status check in the test because it makes the interesting expectations more obvious.

I found that using the create variation of the method was somewhat ambiguous about it's intent and that is better used for intermediary steps where a specific response is needed yet not an important aspect of the test.

This was a realisation I had during the conversion of these tests and so some of the others use the regular create method which is an inconsistency that I could address.

src/test/java/org/folio/moduserstest/ProxyRelationshipsIT.java
marcjohnson-kint: Please can you help me understand better what you mean by `a more rest assured like check`?

How is using the body method that uses a JSON path a better expression of this expectation?

To me, Rest Assured provides a broad set of tools for how to construct requests and check responses. I think the intention of that is to allow the consumers of the library to choose the means they prefer in any given situation.

src/test/java/org/folio/moduserstest/ProxyRelationshipsIT.java
marcjohnson-kint: I think we have different design considerations around the use of rest assured to provide readable tests.

I have no doubt that my rushed conversion of the existing tests means I've made compromises where this could be improved.

This is certainly one of those cases. I was tempted to extract a method into the ProxyRelationships class to hide the stream manipulation.

I think I'm struggling with much of this feedback because I don't understand why relying on JSON path based assertions in RestAssured is better.

I think this is especially the case when the quantity of assertions grows, passing in a map or list of assertions increases the chances that they will get lost in the details and be harder to understand.

To me, checking the status code in the chain of methods from RestAssured is already a compromise as it starts to mix together two separate concerns (making the request and checking the response).

julianladisch (approved):
Thank you for cleaning up this technical dept, well done!

K-Felk (approved):
I've looked through this, and I have to say that I agree with Marc-I find the way the tests are written now to be a lot easier to read than the alternatives presented.  I'd also point out that stylistically, most of the tests on the back end do their assertions standalone at the end of the test, the way they're done here-so there's a consistency argument to be made as well, I think.

PR #247, opened 2022-05-30
Consolidate rest assured usage MODUSERS-309

marcjohnson-kint:
*Purpose*

* Remove duplication of code that uses RestAssured to interact with a record collection API.

*Approach*
* Introduce a new class to consolidate the use of RestAssured to
* Move behaviour to the new record collection client class
* Use new client in existing clients that use rest assured
* Extract a configuration class to allow the common specification and config for RestAssured to be used by non record collection API clients

### DIFF
--- a/src/test/java/org/folio/support/http/OkapiHeaders.java
+++ b/src/test/java/org/folio/support/http/OkapiHeaders.java
@@ -1,10 +1,16 @@
 package org.folio.support.http;
 
+import lombok.AllArgsConstructor;
 import lombok.Value;
 
 @Value
+@AllArgsConstructor
 public class OkapiHeaders {
   String okapiUrl;
   String tenantId;
   String token;
+
+  public OkapiHeaders(OkapiUrl okapiUrl, String tenantId, String token) {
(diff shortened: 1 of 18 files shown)


src/test/java/org/folio/support/http/OkapiHeaders.java
```diff
@@ -1,10 +1,16 @@
...
 @Value
+@AllArgsConstructor
 public class OkapiHeaders {
   String okapiUrl;
   String tenantId;
   String token;
+
+  public OkapiHeaders(OkapiUrl okapiUrl, String tenantId, String token) {
+    this(okapiUrl.toString(), tenantId, token);
+  }
```
kaladay, line 15: I don't understand this part.
Doesn't lombok's `@AllArgsConstructor` create a constructor like this already?
Why have both?

kaladay: Nevermind, I realized after I posted this that the type is `OkapiUrl` and the internal type is a `String`.

kaladay (approved):

PR #248, opened 2022-06-01
MODUSERS-311 Upgrade to RMB 34.0.0.

kaladay:
https://issues.folio.org/browse/MODUSERS-311

Due to a change introduced by [RMB-815](https://issues.folio.org/browse/RMB-815), the org.folio.okapi:okapi-testing is brought in and used for `UtilityClassTester`.

### DIFF
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -178,6 +178,10 @@
           "tOps" : "ADD"
         },
         {
+          "fieldName" : "personal.preferredFirstName",
+          "tOps" : "ADD"
+        },
+        {
           "fieldName" : "personal.email",
           "tOps" : "ADD"
         },
--- a/src/test/java/org/folio/rest/impl/UsersAPIIT.java
+++ b/src/test/java/org/folio/rest/impl/UsersAPIIT.java
@@ -312,6 +312,40 @@ class UsersAPIIT {
   }
 
   @Test
+  void canFindUserByPreferredFirstName() {
+    usersClient.createUser(User.builder()
+      .username("steve")
+      .build());
+
+    usersClient.createUser(User.builder()
+      .username("wilson")
+      .active(true)
+      .personal(Personal.builder()
+        .firstName("wilson")
+        .preferredFirstName("will")
+        .lastName("anderson")
+        .build())
+      .build());
+
+    final var userToFind = User.builder()
+      .username("juliab")
+      .active(true)
+      .personal(Personal.builder()
+        .firstName("julia")
+        .preferredFirstName("jules")
+        .lastName("brockhurst")
+        .build())
+      .build();
+
+    final var userToFindId = usersClient.createUser(userToFind).getId();
+    final var foundUsers = usersClient.getUsers("personal.preferredFirstName==\"jules\"");
+
+    assertThat(foundUsers.getTotalRecords(), is(1));
+    assertThat(foundUsers.getFirstUser().getUsername(), is("juliab"));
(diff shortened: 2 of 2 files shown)


pom.xml
```diff
@@ -143,6 +143,12 @@
       <version>${raml-module-builder.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.folio.okapi</groupId>
+      <artifactId>okapi-testing</artifactId>
+      <version>${okapi-testing.version}</version>
+      <scope>test</scope>
+    </dependency>
```
marcjohnson-kint, line 151: Why was this dependency added?

marcjohnson-kint:
> Due to a change introduced by [RMB-815](https://issues.folio.org/browse/RMB-815), the org.folio.okapi:okapi-testing is brought in and used for UtilityClassTester.

Why does this module need an additional dependency due to an internal change in RMB?

Shouldn't this be handled by a transitory dependency from RMB?

pom.xml
```diff
@@ -499,7 +505,8 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-    <raml-module-builder.version>33.2.4</raml-module-builder.version>
+    <raml-module-builder.version>34.0.0</raml-module-builder.version>
+    <okapi-testing.version>4.14.1</okapi-testing.version>
```
marcjohnson-kint, line 509: Personally, I don't think there is value in extracting an intermediary property for versions that are only used by a single artefact.

K-Felk:
> Why does this module need an additional dependency due to an internal change in RMB?
> 
> Shouldn't this be handled by a transitory dependency from RMB?

I suspect you will have to ask the RMB folks for an answer-what Kevin's done here is exactly how the upgrade notes tell you to handle this issue-including the new dependency.

kaladay:
> > Due to a change introduced by [RMB-815](https://issues.folio.org/browse/RMB-815), the org.folio.okapi:okapi-testing is brought in and used for UtilityClassTester.
> 
> Why does this module need an additional dependency due to an internal change in RMB?
> 
> Shouldn't this be handled by a transitory dependency from RMB?

[According to the PR](https://github.com/folio-org/raml-module-builder/pull/1019/files#diff-6db77c7e404883ffaf8d563fff39dbcdf6142fe5a95193e927bae5bf7654a15d) (and confirmed by failed attempts to compile without this test dependency import), we now have to have this dependency import because it is no longer part of RMB.

I could investigate removing the use of `UtilityClassTester` in the tests but I prefer to avoid altering tests when trying to perform a dependency upgrade such as this.

marcjohnson-kint:
> [According to the PR](https://github.com/folio-org/raml-module-builder/pull/1019/files#diff-6db77c7e404883ffaf8d563fff39dbcdf6142fe5a95193e927bae5bf7654a15d) (and confirmed by failed attempts to compile without this test dependency import), we now have to have this dependency import because it is no longer part of RMB.

> I suspect you will have to ask the RMB folks for an answer-what Kevin's done here is exactly how the upgrade notes tell you to handle this issue-including the new dependency.

Thanks. I think I misunderstood this change. I thought the dependency was being added in order to make RMB work, not that the code that is relied upon has been moved.

> I could investigate removing the use of UtilityClassTester in the tests but I prefer to avoid altering tests when trying to perform a dependency upgrade such as this.

I think we should do this. It's utility (which I consider very small) does not warrant the additional dependency.

I agree that we shouldn't do it as part of this change.

marcjohnson-kint (approved):

PR #249, opened 2022-06-02
MODUSERS-310 Add full text search index for perferredFirstName.

kaladay:
https://issues.folio.org/browse/MODUSERS-310

### DIFF
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -178,6 +178,10 @@
           "tOps" : "ADD"
         },
         {
+          "fieldName" : "personal.preferredFirstName",
+          "tOps" : "ADD"
+        },
+        {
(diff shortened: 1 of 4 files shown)


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -177,6 +177,10 @@
           "fieldName" : "personal.lastName",
           "tOps" : "ADD"
         },
+        {
+          "fieldName" : "personal.preferredFirstName",
+          "tOps" : "ADD"
+        },
```
marcjohnson-kint, line 183: Is there a test for searching by preferred first name?

If there isn't, please add one.

kaladay: There is one now.

marcjohnson-kint: Thank you.

PR #253, opened 2022-06-16
Remove argument captor MODUSERS-317

marcjohnson-kint:
*Purpose*

Removing warnings related to the use of argument capture. The warnings are irritating and the approach makes the tests harder to follow as the arrangement of these expectations are done _after_ the production code is executed.

As part of this work I discovered that there were _no tests_ for successful expiration of users.

*Approach*
* Add integration tests for disabling expired users
* Change the existing tests to use `doAnswer` in order to arrange expectations prior to execution
* Change the existing tests and production code to use future rather than handler based methods from PostgresClient

### DIFF
--- a/src/test/java/org/folio/rest/impl/UsersAPITest.java
+++ b/src/test/java/org/folio/rest/impl/UsersAPITest.java
@@ -1,30 +1,38 @@
 package org.folio.rest.impl;
 
-import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.CoreMatchers.instanceOf;
+import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.CoreMatchers.startsWith;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.mockito.Mockito.*;
+import static org.mockito.Mockito.any;
+import static org.mockito.Mockito.anyBoolean;
+import static org.mockito.Mockito.anyString;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.mock;
 
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
+
 import javax.ws.rs.core.Response;
+
 import org.folio.cql2pgjson.exception.FieldException;
 import org.folio.rest.jaxrs.model.Address;
 import org.folio.rest.jaxrs.model.AddressType;
+import org.folio.rest.jaxrs.model.CustomFields;
 import org.folio.rest.jaxrs.model.Personal;
 import org.folio.rest.jaxrs.model.User;
-import org.folio.rest.jaxrs.model.CustomFields;
-import org.folio.rest.persist.PostgresClient;
 import org.folio.rest.persist.Criteria.Criterion;
+import org.folio.rest.persist.PostgresClient;
 import org.folio.rest.persist.interfaces.Results;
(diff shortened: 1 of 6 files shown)


src/test/java/org/folio/moduserstest/ExpirationIT.java
```diff
@@ -56,9 +60,45 @@ public void beforeEach() {
...
+      .active(true)
+      .expirationDate(ZonedDateTime.now().plusHours(3))
+      .build());
+
+    expirationClient.attemptToTriggerExpiration("diku")
+      .statusCode(is(HTTP_NO_CONTENT));
+
+    final var fetchedUser = usersClient.getUser(unexpiredUser.getId());
+
+    assertThat(fetchedUser.getActive(), is(true));
```
julianladisch, line 101: The tests should not depend on the time zone of the machine that runs the tests:

```
sudo timedatectl set-timezone US/Hawaii
mvn clean install
```

```
[ERROR] Failures:
[ERROR]   ExpirationIT.expiredUsersAreDisabled:81
Expected: is <false>
     but: was <true>
```

```
sudo timedatectl set-timezone Pacific/Auckland
mvn clean install
```

```
[ERROR] Failures:
[ERROR]   ExpirationIT.unexpiredUsersIsNotDisabled:101
Expected: is <true>
     but: was <false>
```

marcjohnson-kint: @julianladisch Thank you for reporting this.

I'll investigate further as the time zone that is used to set the expiration date should be irrelevant. The user should expire regardless.

marcjohnson-kint: I don't think I'm able to recreate this. Do you know of a platform agnostic way to do this e.g. by setting the time for the JVM instance that runs for maven?

And I wonder why it didn't fail on the build server which also uses a different time zone.

marcjohnson-kint: I think I've managed to recreate the issue. I think it is broader than the work in this pull request and will write up an issue describing what I've discovered.

I really appreciate you trying this experiment and reporting this, I think it's uncovered something potentially important.

marcjohnson-kint: Actually, scratch what I said above.

I'm going to file the issue and raise separate pull request that adds these integration tests and fixes this bug specifically targeted at the mainline. That way these changes aren't mixed together.

src/test/java/org/folio/rest/impl/UsersAPITest.java
```diff
@@ -174,26 +164,41 @@ void checkAddressTypeValidCanHandleException(VertxTestContext context) {
...
+    final var postgresClient = mock(PostgresClient.class);
+
+    when(postgresClient.get(anyString(), any(), any(Criterion.class), anyBoolean()))
+      .thenReturn(result);
+
+    final var future = new UsersAPI()
+      .checkAddressTypeValid("someAddressTypeId", postgresClient);
+
+    future.onComplete(context.failing(e ->
+      context.verify(() -> handler.handle(future.cause()))));
```
julianladisch, line 201:
```
    future.onComplete(context.failing(e -> handler.handle(e)));
```
or
```
    future.onComplete(context.failing(handler::handle));
```
context.failing has context.verify included, therefore context.verify is not needed.
context.failing provides the value of future.cause() as e.

julianladisch (approved):
Looks good, thanks!

kaladay (approved):

PR #257, opened 2022-07-07
Incorrect query for expired users MODUSERS-318

marcjohnson-kint:
During some [recent development work](https://github.com/folio-org/mod-users/pull/253#pullrequestreview-1008861334) on mod-users, [~julianladisch] identified that some of the tests failed in a non-UTC timezone.

If mod-users runs in a non-UTC timezone, expiration can either occur too early or be delayed, by roughly the hours offset from UTC.

This is caused by the current implementation of deactivation of expired users assumes that mod-users will be run in a UTC timezone and thus assumed that the current date / time could be used without being formatted as UTC.

This was found in that other work because there were no tests for expiration of users previously.

### DIFF
--- a/src/main/java/org/folio/rest/utils/ExpirationTool.java
+++ b/src/main/java/org/folio/rest/utils/ExpirationTool.java
@@ -1,11 +1,11 @@
 package org.folio.rest.utils;
 
 
+import static java.time.format.DateTimeFormatter.ISO_INSTANT;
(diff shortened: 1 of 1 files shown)


julianladisch (approved):
Looks good, thanks!

marcjohnson-kint:
> Looks good, thanks!

Thank you for discovering the issue in the first place

K-Felk (approved):

PR #258, opened 2022-07-13
Improve coverage MODUSERS-319

marcjohnson-kint:
## Motivation

Recent changes have reduced the code coverage in mod-users. I wanted to try to improve that metric and fill in existing gaps.

## Scope

Includes improvements to:
* Users API
* Groups API
* Address Types API
* Proxies for API
* Patron Pin API

Further improvements, including to the following APIs require more substantial changes and will be considered for future pull requests:
* Departments API (because the integration tests for this API need to be improved as well)
* Patron Pin API (further improvement requires reworking the flow of the code)
* Users API (further improvement requires reworking the flow of the code)

## Approach

* Introduce a failure handler to reduce duplication of handling unexpected failures and to specifically test that behaviour
* Extract out database operations to repositories to make testing them in isolation and faking failures easier (though mocking the PostgresClient is not good practice and should eventually be removed)
* Extract out operations that may fail into services
* Add missing tests for important behaviour that wasn't covered e.g. unknown address type and duplicate barcode when updating records
* Removed redundant code that seems to have carried over from when error checks were performed a different way


src/main/java/org/folio/rest/impl/ProxiesForAPI.java
```diff
@@ -49,21 +46,24 @@ public void postProxiesfor(String lang, ProxiesFor entity,
...
+      PostProxiesforResponse::respond500WithTextPlain);
+
     userAndProxyUserComboExists(entity, postgresClient)
-      .onFailure(handleFailure(asyncResultHandler))
       .onSuccess(proxyAlreadyExists -> {
         if (Boolean.TRUE.equals(proxyAlreadyExists)) {
           logger.error("Proxy relationship already exists: {}", entity.getId());
           Errors existsError = ValidationHelper.createValidationErrorMessage(
-            "proxyFor", entity.getId(), "Proxy relationship already exists");
+          "proxyFor", entity.getId(), "Proxy relationship already exists");
```
julianladisch, line 57: Why has this indentation been removed?

src/main/java/org/folio/service/impl/RecordRepositoryImpl.java
```diff
@@ -87,14 +87,14 @@ public Future<List<User>> findUsersByFieldValues(RecordUpdate recordUpdate, Stri
...
-  public Future<Boolean> updateUser(User user, String tenantId) {
+  public Future<Void> updateUser(User user, String tenantId) {
     user.setUpdatedDate(new Date());
 
     Promise<RowSet<Row>> promise = Promise.promise();
 
     pgClient(tenantId).update(USERS_TABLE, user, user.getId(), promise);
 
-    return promise.future().map(updateResult -> updateResult.size() == 1);
+    return promise.future().map(r -> null);
```
julianladisch, line 97: return pgClient(tenantId).update(USERS_TABLE, user, user.getId()).mapEmpty();

src/main/java/org/folio/rest/impl/ProxiesForAPI.java
marcjohnson-kint: That is a mistake. Thanks for finding that. I will correct it.

src/main/java/org/folio/service/impl/RecordRepositoryImpl.java
marcjohnson-kint: Thanks. I've used `mapEmpty` before and forgot about it, I will change this.

julianladisch (approved):
Looks good, thanks!

src/main/java/org/folio/rest/impl/ProxiesForAPI.java
```diff
@@ -49,21 +46,26 @@ public void postProxiesfor(String lang, ProxiesFor entity,
...
+
           asyncResultHandler.handle(Future.succeededFuture(
             PostProxiesforResponse.respond422WithApplicationJson(existsError)));
           return;
         }
 
         PgUtil.post(PROXY_FOR_TABLE, entity, okapiHeaders, vertxContext,
-          PostProxiesforResponse.class, asyncResultHandler);
-      });
+        PostProxiesforResponse.class, asyncResultHandler);
```
kaladay, line 66: Shouldn't this line be tabbed over?
It is a continuation of the function parameters from line 65.

src/main/java/org/folio/service/impl/PatronPinService.java
```diff
@@ -0,0 +1,34 @@
...
+
+  private final PasswordHashService passwordHashService;
+
+  public PatronPinService(PasswordHashService passwordHashService) {
+    this.passwordHashService = passwordHashService;
+  }
+
+  public String derivePin(String pin, String userId) {
+    try {
+      // Should the salt also include the tenant?
```
kaladay, line 24: Salts should always, in theory, be random (but saved somewhere) so that they act to further obscure a hash.
By adding a known, such as a username or in this case a tenant, it is rather self-defeating.
The theoretical attacker would easly be able to know the "tenant" part, thereby reducing the strength of the salt.

For example a salt of `12bob`, with `bob` being the tenant, would effectively be the same has having a salt of `12`.

With that said, having two knowns (userId and Tenant) is probably more secure than one known (userId).
At least the ordering can offer some slight security.

I would recommend (somehow) having the salt generated and saved, such as a configuration file that loads the salt and then at some point this random salt is passed to this function. I believe there are some implementations that store this data in the database rather than on the disk. The trick is figuring out the safest way to store the salt while still being able to fetch and use it in a way that works with a given project. More secure salts would have a different salt per user, but this approach would require more storage and fetching.

src/main/java/org/folio/service/impl/PatronPinService.java
```diff
@@ -0,0 +1,34 @@
...
+public class PatronPinService {
+  private static final Logger logger = LogManager.getLogger(PatronPinService.class);
+
+  private final PasswordHashService passwordHashService;
+
+  public PatronPinService(PasswordHashService passwordHashService) {
+    this.passwordHashService = passwordHashService;
+  }
+
+  public String derivePin(String pin, String userId) {
```
kaladay, line 22: You should probably rename `String userId` to `String salt`.

kaladay (changes requested):

src/main/java/org/folio/service/impl/PatronPinService.java
marcjohnson-kint: I chose to refer to this as the userId in order to inform the client of the code that it is the `user ID` that is expected, if it is named `salt`, the client has to guess what the expected value is.

I consider the use of the user ID as the salt a decision made within this class. The client does not need to know that is how the user ID is used.

src/main/java/org/folio/service/impl/PatronPinService.java
marcjohnson-kint: TLDR: I purposely didn't change the decision that had already been made to use the user ID as the salt.

I think feedback about whether this is an appropriate salt would be better given to @ianibo and the Leipzig folks who asked for this functionality, the original issues for this work can be found [here|https://issues.folio.org/browse/MODUSERS-254].

This code could be switched to use a secure random number generator and store the salt with the pin.

It would mean needing to migrate the existing data during an upgrade. And due to limitations with how the data model and API record representation are the same in RMB modules, would need some work to make sure it wasn't exposed to a client.

> Salts should always, in theory, be random (but saved somewhere) so that they act to further obscure a hash. By adding a known, such as a username or in this case a tenant, it is rather self-defeating. The theoretical attacker would easly be able to know the "tenant" part, thereby reducing the strength of the salt.

I agree that a cryptographically random number could be a better choice of salt than a known value.

My cryptography is pretty rusty, based upon this [article|https://auth0.com/blog/adding-salt-to-hashing-a-better-way-to-store-passwords/] I think the two goals of salting (getting a different hash for the same pin and not being able to use a shared table for table attacks) are still achieved by using a different hash for each user, even if it isn't the optimal solution (because it could allow for different password / pin fields with the same value, to end up with the same hash, which could reveal information to an attacker.

> For example a salt of `12bob`, with `bob` being the tenant, would effectively be the same has having a salt of `12`.
> 
> With that said, having two knowns (userId and Tenant) is probably more secure than one known (userId). At least the ordering can offer some slight security.
> 
> I would recommend (somehow) having the salt generated and saved, such as a configuration file that loads the salt and then at some point this random salt is passed to this function. I believe there are some implementations that store this data in the database rather than on the disk. The trick is figuring out the safest way to store the salt while still being able to fetch and use it in a way that works with a given project. More secure salts would have a different salt per user, but this approach would require more storage and fetching.

As I understand it, salts are also typically intended to be different for each key, as this contributes to

src/main/java/org/folio/service/impl/PatronPinService.java
kaladay: ok

ianibo (approved):

kaladay (approved):

PR #269, opened 2023-02-03
MODUSERS-338: adding full-text index for custom fields

K-Felk:


marcjohnson-kint (approved):
Did the addition to schema.json match up with the manually created index referred to in the issue?

K-Felk:
> Did the addition to schema.json match up with the manually created index referred to in the issue?

It looks as though it does.  Both seem to attach an index with identical names to the users table.  Properties look the same as well.

marcjohnson-kint:
> > Did the addition to schema.json match up with the manually created index referred to in the issue?
> 
> It looks as though it does. Both seem to attach an index with identical names to the users table. Properties look the same as well.

Thanks for checking

PR #280, opened 2023-03-25
MODUSERS-335: Upgrading tenant resets reference and sample records

julianladisch:
https://issues.folio.org/browse/MODUSERS-335

When upgrading the mod-users version for a tenant only install or restore data that has changed.


src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -26,16 +27,35 @@ Future<Integer> loadData(TenantAttributes attributes, String tenantId,
...
 
           tl.withKey("loadSample").withLead("sample-data");
-          tl.withIdContent();
-          tl.add("users");
+          if (isNew(attributes, "15.4.0")) {
+            tl.withIdContent().add("users-15.4.0", "users");
+          }
+          if (isNew(attributes, "17.3.0")) {
+            tl.withIdContent().add("users-17.3.0", "users");
+          }
```
steveellis, line 43: Can you provide any additional details about how the TenantRefAPI is used in upgrading? How do these changes reduce extra work for staff? Will this break upgrades?

steveellis (approved):

PR #282, opened 2023-03-31
MODUSERS-351: Release 19.1.1

julianladisch:
https://issues.folio.org/browse/MODUSERS-351


marcjohnson-kint (commented):
I don't think I'm in a position to do a sensible review of these changes, as I am unfamiliar with the mechanisms that have been put in place.

I don't think a approval from me is required for this release to go through.

PR #283, opened 2023-04-07
[MODUSERS-347] Implement sending USER_CREATED event using transactional outbox pattern

Abdulkhakimov:
Transactional outbox pattern used for sending USER_CREATED event.
This event will be consumed by mod-consortia to store primary affiliation of the user in local persistence.
Detailed diagram is here:
image
In general, this implementation of transactional outbox pattern has been approved by the FOLIO Tech Council: https://wiki.folio.org/display/DD/Implementing+the+Transactional+Outbox+pattern


grabsefx (approved):

src/main/java/org/folio/event/service/UserOutboxService.java
```diff
@@ -0,0 +1,121 @@
...
+
+        logger.info("Fetched {} event logs from outbox table, going to send them to kafka", logs.size());
+        List<Future<Boolean>> futures = getKafkaFutures(logs, okapiHeaders);
+        return GenericCompositeFuture.join(futures)
+          .map(logs.stream().map(OutboxEventLog::getEventId).collect(Collectors.toList()))
+          .compose(eventIds -> {
+            if (CollectionUtils.isNotEmpty(eventIds)) {
+              return outboxRepository.deleteBatch(conn, eventIds, tenantId)
+                .onSuccess(rowsCount -> logger.info("{} logs have been deleted from outbox table", rowsCount))
+                .onFailure(ex -> logger.error("Logs deletion filed", ex));
```
tatsiana-tarhonskaya, line 69: Logs deletion failed

src/main/java/org/folio/repository/ConsortiaEventsLogRepository.java
```diff
@@ -0,0 +1,86 @@
...
+
+public class ConsortiaEventsLogRepository {
+
+  private static final String OUTBOX_TABLE_NAME = "outbox_event_log";
+  private static final String EVENT_ID_FIELD = "event_id";
+  private static final String ENTITY_TYPE_FIELD = "entity_type";
+  private static final String ACTION_FIELD = "action";
+  private static final String PAYLOAD_FIELD = "payload";
+  private static final String INSERT_SQL = "INSERT INTO %s.%s (event_id, entity_type, action, action_date, payload) VALUES ($1, $2, $3, $4, $5)";
+  private static final String SELECT_EVENT_LOGS = "SELECT * FROM %s.%s ORDER BY action_date LIMIT 1000";
```
tatsiana-tarhonskaya, line 28: Should an index for action_date be created probably?

src/main/java/org/folio/event/service/UserOutboxService.java
Abdulkhakimov: thank you, fixed

src/main/java/org/folio/repository/ConsortiaEventsLogRepository.java
Abdulkhakimov: Thanks, added

src/test/java/org/folio/rest/impl/UsersAPIIT.java
```diff
@@ -60,6 +69,13 @@ static void beforeAll(Vertx vertx, VertxTestContext context) {
     final var okapiUrl = new OkapiUrl("http://localhost:" + port);
     final var headers = new OkapiHeaders(okapiUrl, tenant, token);
 
+    kafkaCluster = EmbeddedKafkaCluster.provisionWith(defaultClusterConfig());
+    kafkaCluster.start();
+    String[] hostAndPort = kafkaCluster.getBrokerList().split(":");
+    System.setProperty(KAFKA_HOST, hostAndPort[0]);
+    System.setProperty(KAFKA_PORT, hostAndPort[1]);
+    System.setProperty(KAFKA_ENV, KAFKA_ENV_VALUE);
```
steveellis, line 77: Seeing these lines are repeated quite a bit in the tests. Maybe move them out to a static helper class to make it more DRY.

SinghAdes (approved):

src/test/java/org/folio/moduserstest/ProxyRelationshipsIT.java
```diff
@@ -64,8 +80,10 @@ public static void beforeAll(Vertx vertx, VertxTestContext context) {
   }
 
   @BeforeEach
-  public void beforeEach() {
+  public void beforeEach(VertxTestContext context) {
     proxiesClient.deleteAllProxies();
+
+    context.completeNow();
```
grabsefx, line 86: since there is only synchronous code in this block I dont see the reason why test context should be used here

src/test/java/org/folio/moduserstest/ExpirationIT.java
```diff
@@ -107,4 +124,10 @@ void cannotTriggerExpirationForUnknownTenant() {
     expirationClient.attemptToTriggerExpiration("made-up-tenant")
       .statusCode(is(HTTP_INTERNAL_ERROR));
   }
+
+  @AfterAll
+  public static void after(VertxTestContext context) {
+    context.completeNow();
```
grabsefx, line 130: I think context completion should be performed in the end, after all operations finished. Otherwise this could hide a bigger problem. 
Moreover, I didnt find any removal of created tenants at the end of tests. Probably this would be a good verification of all async handlers completed and all database transactions finished

ramls/userConsortiaEvent.json
```diff
@@ -0,0 +1,44 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "User consortia event",
```
marcjohnson-kint, line 3: What is a user consortia event?

ramls/users.raml
```diff
@@ -10,6 +10,8 @@ documentation:
 types:
   user: !include userdata.json
   userdataCollection: !include userdataCollection.json
+  outboxEventLog: !include outboxEventLog.json
+  userConsortiaEvent: !include userConsortiaEvent.json
```
marcjohnson-kint, line 14: Why are these in the RAML definitions when there is no endpoints associated with either of these types?

src/main/java/org/folio/event/UserConsortiaEventType.java
```diff
@@ -0,0 +1,15 @@
+package org.folio.event;
+
+public enum UserConsortiaEventType {
```
SerhiiNosko, line 3: Please remove Consortia from each java class name, not only mod-consortia can consume these events in future, so lets make names more generic

raman-auramau:
> > In general, this implementation of transactional outbox pattern has been approved by the FOLIO Tech Council: https://wiki.folio.org/display/DD/Implementing+the+Transactional+Outbox+pattern
> 
> This does not fit with my understanding. As I understand it, the [latest decision](https://wiki.folio.org/display/TC/DR-000004+-+Cross+application+data+sync+proposal) made by the TC in this area excluded the transactional outbox pattern from the scope of the decision. Approving that extension has been [waiting](https://wiki.folio.org/display/TC/2022-08-10+Meeting+notes) on a proof of concept implementation.

@marcjohnson-kint , thanks for pointing out. That's right, the mentioned TC solution [DR-000004 - Cross application data sync proposal](https://wiki.folio.org/display/TC/DR-000004+-+Cross+application+data+sync+proposal) clearly says that "_the use of the Outbox and Inbox patterns in the context of the solution described above is out of scope of this decision. The Technical Council may choose to revisit this at a later date._" Since there was no such revision, the statement about "_this implementation of transactional outbox pattern has been approved by the FOLIO Tech Council_" is incorrect.

However, do you think that since the Thunderjet team implemented the Outbox pattern in the code (here and as a part of Acquisition Event Log), this could be the basis to request a revision by the Tech Consul?

marcjohnson-kint:
> do you think that since the Thunderjet team implemented the Outbox pattern in the code (here and as a part of Acquisition Event Log), this could be the basis to request a revision by the Tech Consul?

Yes, I think these implementations could be used as the basis for a follow up proposal.

I think the TC will likely ask questions about performance and reliability. I don't know what work has been done in these areas e.g. knowing what the realistic delay in publishing a message is.

ramls/userConsortiaEvent.json
Abdulkhakimov: Renamed

src/test/java/org/folio/moduserstest/ExpirationIT.java
Abdulkhakimov: Rewritten by moving initialization logic to abstract class

src/test/java/org/folio/moduserstest/ProxyRelationshipsIT.java
Abdulkhakimov: removed

src/test/java/org/folio/rest/impl/UsersAPIIT.java
Abdulkhakimov: Moved to another class

SerhiiNosko (approved):

ramls/users.raml
Abdulkhakimov: If you don't specify a JSON schema in the RAML definitions, it won't be generated. I thought that user.raml is the most appropriate place.

marcjohnson-kint: Ok, this is a workaround to generate classes for the event log and user event types?

Abdulkhakimov: You're partially correct. However, outboxEventLog is an entity for storing payloads for the transactional outbox pattern for any types of events, while userEvent is the payload that will be sent to Kafka. I created a separate enum as a Java class without generating it for event types.

marcjohnson-kint: > However, outboxEventLog is an entity for storing payloads for the transactional outbox pattern for any types of events, while userEvent is the payload that will be sent to Kafka. 

How does that distinction affect the code generation?

Abdulkhakimov: we can't reference the userEvent directly in the outboxEventLog fields, we store the userEvent in a json representation in the payload field of the outboxEvenLog. This is done to reuse outboxEventLog for other entities.

marcjohnson-kint: Sure, I understand that. I don't understand how that affects the code generation. I don't think it's important, I now understand the desire for the code generation and this workaround was considered the best way.

grabsefx (approved):

descriptors/ModuleDescriptor-template.json
```diff
@@ -236,6 +236,12 @@
           "pathPattern" : "/users/expire/timer",
           "unit" : "minute",
           "delay" : "1"
+        },
+        {
+          "methods" : [ "POST" ],
+          "pathPattern" : "/users/users-outbox/process",
```
marcjohnson-kint, line 242: What is the significance of the second `users` in this path?

Is there intended to be a separate outbox for each record type in the future?

descriptors/ModuleDescriptor-template.json
```diff
@@ -510,7 +516,9 @@
       { "name": "DB_DATABASE", "value": "okapi_modules" },
       { "name": "DB_QUERYTIMEOUT", "value": "60000" },
       { "name": "DB_CHARSET", "value": "UTF-8" },
-      { "name": "DB_MAXPOOLSIZE", "value": "5" }
+      { "name": "DB_MAXPOOLSIZE", "value": "5" },
+      { "name": "KAFKA_HOST", "value": "10.0.2.15"},
+      { "name": "KAFKA_PORT", "value": "9092" }
```
marcjohnson-kint, line 521: Introducing Kafka as a new dependency to the module means that this version of the module cannot be used with the same infrastructure dependencies as previous versions. 

I think that means the major version of the module should be changed (and news / release notes provided) to communicate that change.

ramls/examples/outboxEventLog.sample
```diff
@@ -0,0 +1,24 @@
+{
```
marcjohnson-kint, line 1: Where is this example used?

As I understand it, these JSON schema are only defined in order to trigger code generation, and thus are not defined in any endpoint definitions, which is where examples would typically be used.

ramls/userEvent.json
```diff
@@ -0,0 +1,44 @@
+{
```
marcjohnson-kint, line 1: This defines a different structure for the event message than the [original pattern](https://github.com/folio-org/mod-inventory-storage/blob/f40481bf2cbf5a1385141e5169e633b8c15a83d6/src/main/java/org/folio/services/domainevent/DomainEvent.java) established in mod-inventory-storage when the Falcon team did the inventory search work.

What is the reasoning for using a different structure?

ramls/outboxEventLog.json
```diff
@@ -0,0 +1,33 @@
...
+      "description": "Entity type of the event",
+      "type": "string",
+      "enum": [
+        "User"
+      ]
+    },
+    "action": {
+      "description": "Action for entity",
+      "type": "string"
+    },
```
marcjohnson-kint, line 21: Why is there an action defined in the outbox event log and the user event?

src/main/java/org/folio/rest/impl/UsersAPI.java
```diff
@@ -228,40 +238,47 @@ public void postUsers(String lang, User entity, RoutingContext routingContext,
...
+        if (isDuplicateBarcodeError(reply)) {
+          asyncResultHandler.handle(
+            succeededFuture(PostUsersResponse.respond422WithApplicationJson(
+              ValidationHelper.createValidationErrorMessage(
+                "barcode", entity.getBarcode(),
+                "This barcode has already been taken"))));
+          return;
+        }
+        logger.debug("Save successful");
+        userOutboxService.processOutboxEventLogs(vertxContext.owner(), okapiHeaders);
```
marcjohnson-kint, line 279: My interpretation of this code is that it processes the outbox at the end of handling the API request to create a user.

It is my understanding that the intention of the outbox pattern is that messages are sent separately in a durable process. This seems counter to that pattern. What is the reason for trigger the processing of the outbox during each invocation of the create user API?

src/main/java/org/folio/rest/impl/UsersAPI.java
```diff
@@ -228,40 +238,47 @@ public void postUsers(String lang, User entity, RoutingContext routingContext,
...
-              "This barcode has already been taken"))));
-        return;
-      }
-      logger.debug("Save successful");
-      asyncResultHandler.handle(reply);
-    });
+    String userId = StringUtils.defaultIfBlank(entity.getId(), UUID.randomUUID().toString());
+
+    pgClient.withTrans(conn -> conn.saveAndReturnUpdatedEntity(TABLE_NAME_USERS, userId, entity.withId(userId))
+      .compose(user -> userOutboxService.saveUserOutboxLog(conn, user, UserEvent.Action.CREATE, okapiHeaders)
```
marcjohnson-kint, line 250: How does the transaction get committed?

Is RMB doing something special to know that everything that happens on that connection should be in the same transaction?

src/test/java/org/folio/rest/impl/ReferenceAndSampleDataIT.java
```diff
@@ -6,56 +6,33 @@
...
 import io.vertx.junit5.Timeout;
 import io.vertx.junit5.VertxExtension;
-import io.vertx.junit5.VertxTestContext;
-import lombok.SneakyThrows;
 
 // Loading the reference and sample data can take a long time
 @Timeout(value = 30, timeUnit = SECONDS)
 @ExtendWith(VertxExtension.class)
-class ReferenceAndSampleDataIT {
+class ReferenceAndSampleDataIT extends AbstractRestTest {
```
marcjohnson-kint, line 23: Should messages to be sent to Kafka for users created when sample data is loaded?

marcjohnson-kint (commented):
I couldn't find any tests for the messages being published to Kafka. Where might those be found?

descriptors/ModuleDescriptor-template.json
Abdulkhakimov: Updated NEWS.md

ramls/examples/outboxEventLog.sample
Abdulkhakimov: You're right. Removed the sample.

src/main/java/org/folio/rest/impl/UsersAPI.java
Abdulkhakimov: This is done to send changes to Kafka immediately after saving. If an error occurs during sending, the processOutboxEvent will be called from the API request handler.

src/main/java/org/folio/rest/impl/UsersAPI.java
Abdulkhakimov: Under the hood, the withTrans mechanism is implemented to provide rollback functionality.

src/test/java/org/folio/rest/impl/ReferenceAndSampleDataIT.java
SerhiiNosko: As for me, it is correct, because we should not differ sample data users and real users, even if mod-consortia will be disabled - in this case events for sample data users will stay in the topic until expiration period finishes, but it is OK

okolawole-ebsco: When a user, any user is created in mod-users, the appropriate event should be sent. We should not differentiate between sample and non-sample users.

Abdulkhakimov:
> I couldn't find any tests for the messages being published to Kafka. Where might those be found?

I've added tests

SerhiiNosko:
It seams all comments are resolved and already have 5 approvals, so going to merge this PR, because this actually blocks 4 stories in current sprint, if you have some additional thoughts or questions please post them and we will address in scope of subsequent PRs, we are going to create 2 new PRs in this sprint for mod-users: to send USER_DELETED event  and to consume kafka events mod-consortia.

src/test/java/org/folio/rest/impl/ReferenceAndSampleDataIT.java
marcjohnson-kint: If this is important behaviour, maybe that could be checked in these tests

src/main/java/org/folio/rest/impl/UsersAPI.java
marcjohnson-kint: I can understand how it's useful for that to be done eagerly in order to reduce potential lag for publishing the message. 

I think that somewhat contradicts how the outbox pattern is intended to work. In this case, it presents a challenge with knowing that the background processing is working. 

Given that the message is published during the API request, how do folks know that the background outbox publishing process is working?

ramls/userEvent.json
marcjohnson-kint: > First and foremost, the action type is missing there. We plan to add support for DELETE as well. 

The action type is not missing in that structure. The `type` property represents that type. `DELETE` is one of the types used - https://github.com/folio-org/mod-inventory-storage/blob/f40481bf2cbf5a1385141e5169e633b8c15a83d6/src/main/java/org/folio/services/domainevent/DomainEvent.java#L82

PR #289, opened 2023-05-23
[MODUSERS-364] Create Post user-tenant endpoint

bekhzodkurbonboev:
Resolves [MODUSERS-364](https://issues.folio.org/browse/MODUSERS-364) Create Post user-tenant endpoint
We need this for new approach in mod-authtoken
- After adding some institutional tenant to Consortium mod-consortia will invoke POST endpoint implemented here to create some dummy user tenant record
- Mod-authtoken will call GET /user-tenant endpoint in the institutional tenant to check totalRecords(in case when tenant in header not matches with tenant in token). If totalRecords > 0 - mod-authtoken will allow the cross-tenant request

### DIFF
--- a/src/main/java/org/folio/rest/impl/UserTenantsAPI.java
+++ b/src/main/java/org/folio/rest/impl/UserTenantsAPI.java
@@ -5,6 +5,7 @@ import io.vertx.core.Context;
 import io.vertx.core.Handler;
 import org.apache.commons.lang3.StringUtils;
 import org.folio.event.service.UserTenantService;
+import org.folio.rest.jaxrs.model.UserTenant;
 import org.folio.rest.jaxrs.resource.UserTenants;
 import org.folio.rest.persist.Criteria.Criteria;
 import org.folio.rest.persist.Criteria.Criterion;
@@ -41,6 +42,18 @@ public class UserTenantsAPI implements UserTenants {
         succeededFuture(GetUserTenantsResponse.respond200WithApplicationJson(res))))
       .onFailure(cause -> asyncResultHandler.handle(
         succeededFuture(GetUserTenantsResponse.respond500WithTextPlain(cause.getMessage()))));
+  }
+
+  @Override
+  public void postUserTenants(String lang, UserTenant userTenant, Map<String, String> okapiHeaders,
+                              Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
+    String okapiTenantId = TenantTool.tenantId(okapiHeaders);
+
+    userTenantService.saveUserTenant(userTenant, okapiTenantId, vertxContext.owner())
+      .onSuccess(res -> asyncResultHandler.handle(
+        succeededFuture(Response.noContent().build())))
+      .onFailure(cause -> asyncResultHandler.handle(
+        succeededFuture(PostUserTenantsResponse.respond500WithTextPlain(cause.getMessage()))));
   }
 
   private void addWhereClauseArgumentsToCriterion(String userId, String username, String tenantId, Criterion criterion) {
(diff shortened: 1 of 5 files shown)


src/main/java/org/folio/rest/impl/UserTenantsAPI.java
```diff
@@ -43,6 +44,18 @@ public void getUserTenants(String userId, String username, String tenantId, int
...
   }
 
+  @Override
+  public void postUserTenants(String lang, UserTenant userTenant, Map<String, String> okapiHeaders,
+                              Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
+    String okapiTenantId = TenantTool.tenantId(okapiHeaders);
+
+    userTenantService.saveUserTenant(userTenant, okapiTenantId, vertxContext.owner())
+      .onSuccess(res -> asyncResultHandler.handle(
+        succeededFuture(Response.noContent().build())))
```
SerhiiNosko, line 54: we should use 201 Created status

src/main/java/org/folio/rest/impl/UserTenantsAPI.java
```diff
@@ -43,6 +44,18 @@ public void getUserTenants(String userId, String username, String tenantId, int
...
+      .onFailure(cause -> asyncResultHandler.handle(
+        succeededFuture(PostUserTenantsResponse.respond500WithTextPlain(cause.getMessage()))));
```
tatsiana-tarhonskaya, line 56: some logging would be helpful. or is it done somewhere already?

src/main/java/org/folio/rest/impl/UserTenantsAPI.java
```diff
@@ -36,24 +39,42 @@ public void getUserTenants(String userId, String username, String tenantId, int
...
     addWhereClauseArgumentsToCriterion(userId, username, tenantId, criterion);
+    logger.debug("Trying to get user-tenant records with criterion: {}.", criterion);
 
     userTenantService.fetchUserTenants(okapiTenantId, criterion, vertxContext.owner())
-      .onSuccess(res -> asyncResultHandler.handle(
-        succeededFuture(GetUserTenantsResponse.respond200WithApplicationJson(res))))
-      .onFailure(cause -> asyncResultHandler.handle(
-        succeededFuture(GetUserTenantsResponse.respond500WithTextPlain(cause.getMessage()))));
+      .onSuccess(res -> {
+        logger.info("Number of existing user-tenant records: {}.", res.getTotalRecords());
```
SerhiiNosko, line 46: Lets use debug here for get request, because it will be called by mod-authtoken often(mod-authtoken will have a chache for it but nevertheless).

src/main/java/org/folio/rest/impl/UserTenantsAPI.java
```diff
@@ -35,12 +39,42 @@ public void getUserTenants(String userId, String username, String tenantId, int
...
+    userTenantService.saveUserTenant(userTenant, okapiTenantId, vertxContext.owner())
+      .onSuccess(res -> {
+        logger.info("user-tenant with id: {}, userId: {}, userName: {}, tenantId: {} has been saved successfully.",
+          userTenant.getId(), userTenant.getUserId(), userTenant.getUserName(), userTenant.getTenantId());
+        asyncResultHandler.handle(
+          succeededFuture(Response.status(201).build()));
+      })
+      .onFailure(cause -> {
+        String message = cause.getLocalizedMessage();
+        logger.error(message, cause);
```
SerhiiNosko, line 74: lets also use predefined message in both endpoints, because getLocalizedMessage we will see from stacktrace, but if we want search log by some predefined message - it would be easier

src/main/java/org/folio/rest/impl/UserTenantsAPI.java
bekhzodkurbonboev: Changed logging level to debug.

src/main/java/org/folio/rest/impl/UserTenantsAPI.java
bekhzodkurbonboev: Fixed

SerhiiNosko (approved):

tatsiana-tarhonskaya (approved):

PR #295, opened 2023-07-10
[MODUSERS-362]-Added Update event.

SinghAdes:
https://issues.folio.org/browse/MODUSERS-362


src/main/java/org/folio/repository/UserTenantRepository.java
```diff
@@ -32,8 +32,11 @@ public class UserTenantRepository {
   private static final String USER_TENANT_TABLE_NAME = "user_tenant";
   private static final String ID_FIELD = "id";
   private static final String TOTAL_COUNT_FIELD = "total_count";
-  private static final String INSERT_SQL = "INSERT INTO %s.%s (id, user_id, username, tenant_id, creation_date) VALUES ($1, $2, $3, $4, $5)";
+  private static final String INSERT_SQL = "INSERT INTO %s.%s (id, user_id, username, tenant_id, creation_date, central_tenant_id, email, mobile_phone_number, phone_number)" +
+    " VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9)";
   private static final String DELETE_SQL = "DELETE FROM %s.%s WHERE user_id = $1 AND tenant_id = $2";
+  private static final String UPDATE_SQL = "UPDATE %s.%s SET central_tenant_id = $1, email = $2, mobile_phone_number = $3, phone_number = $4 WHERE user_id = $5";
```
SerhiiNosko, line 38: username is also updatable field, central_tenant_id - not

src/main/java/org/folio/verticle/consumers/ConsortiumUpdateEventsHandler.java
```diff
@@ -0,0 +1,54 @@
...
+    OkapiConnectionParams okapiConnectionParams = new OkapiConnectionParams(KafkaHeaderUtils.kafkaHeadersToMap(kafkaHeaders), vertx);
+    UserTenant event = new JsonObject(record.value()).mapTo(UserTenant.class);
+    logger.info("Trying to update of user primary affiliation event with event id: {} for user id: {} with tenant id: {}",
+      event.getId(), event.getUserId(), event.getTenantId());
+
+    userTenantService.updateUserTenant(event, okapiConnectionParams.getTenantId(), okapiConnectionParams.getVertx())
+      .onSuccess(ar -> {
+        logger.info("User primary affiliation event with event id: {} has been updated for user id: {} with tenant id: {}",
+          event.getId(), event.getUserId(), event.getTenantId());
+        result.complete(event.getId());
```
SerhiiNosko, line 44: When possible for such simple cases we should not use promise, so you can use ConsortiumDeleteEventsHandler as a reference. 
Also please adjust logs:
"Trying to update of user primary affiliation event with event id" -> "Trying to update of user primary affiliation with event id"
in all places because we are not updating event, we updating affiliation

src/test/java/org/folio/rest/impl/UserTenantIT.java
```diff
@@ -77,6 +78,26 @@ void canDeleteAllUserTenants() {
...
+    new java.util.Timer().schedule(
+      new java.util.TimerTask() {
+        @Override
+        public void run() {
+          UserTenantCollection collection2 = userTenantClient.getAllUsersTenants();
+          Assertions.assertEquals("Test", collection2.getUserTenants().get(2).getEmail());
+          Assertions.assertEquals("1234", collection2.getUserTenants().get(2).getPhoneNumber());
+        }
+      },
+      10000
```
SerhiiNosko, line 97: For unit test 10 sec can be long, also please add asserts for mobilePhoneNumber and username

SerhiiNosko (approved):

src/test/java/org/folio/rest/impl/UserTenantIT.java
```diff
@@ -59,13 +64,15 @@ public void beforeEach() {
   }
 
   @Test
+  @Order(9)
```
SerhiiNosko, line 67: please avoid this annotaiton in unit tests

SinghAdes: fixed

Dmitriy-Butramyou (approved):

gurleenkaurbp (approved):

PR #296, opened 2023-07-21
[MODUSERBL-169] - getUserTenants criterion changed to or

Dmitriy-Butramyou:
[MODUSERBL-169](https://issues.folio.org/browse/MODUSERBL-169)
Before this change /user-tenant endpoint filters by username, userId, ematil etc only using AND operator.
For Forgot password/Forgot username we need to search by mutiple fields using OR operator and we introduced new param to specify operator for filtering. AND is by default

### DIFF
--- a/src/main/java/org/folio/rest/impl/UserTenantsAPI.java
+++ b/src/main/java/org/folio/rest/impl/UserTenantsAPI.java
@@ -87,6 +87,6 @@ public class UserTenantsAPI implements UserTenants {
     fields.entrySet().stream()
       .filter(entry -> StringUtils.isNotBlank(entry.getValue()))
       .map(entry -> new Criteria().addField(entry.getKey()).setOperation("=").setVal(entry.getValue()).setJSONB(false))
-      .forEach(criterion::addCriterion);
+      .forEach(a -> criterion.addCriterion(a, "or"));
   }
 }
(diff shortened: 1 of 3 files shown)


src/main/java/org/folio/rest/impl/UserTenantsAPI.java
```diff
@@ -87,6 +87,6 @@ private void addWhereClauseArgumentsToCriterion(String userId, String username,
     fields.entrySet().stream()
       .filter(entry -> StringUtils.isNotBlank(entry.getValue()))
       .map(entry -> new Criteria().addField(entry.getKey()).setOperation("=").setVal(entry.getValue()).setJSONB(false))
-      .forEach(criterion::addCriterion);
+      .forEach(a -> criterion.addCriterion(a, "or"));
```
SerhiiNosko, line 90: please add new optional param for this, if it skipped - "and" should be used by default and add unit tests

okolawole-ebsco:
Reading the JIRA linked, I could not decipher why this change was necessary. Please add a purpose or approach to the description of this MR.

Also, I see the addition of "or", is there "and" somewhere else?

SerhiiNosko:
> Reading the JIRA linked, I could not decipher why this change was necessary. Please add a purpose or approach to the description of this MR.
> 
> Also, I see the addition of "or", is there "and" somewhere else?

@okolawole-ebsco we will add description and for all futher PRs as well. Basically for this change we need filtering by OR for forgot password/forgot username cases , because that logic makes search by mutiple fields: username, phoneNumber, mobilePhoneNumber, email.
For othe cases we need AND that will be used by default if parameter missed

tatsiana-tarhonskaya (approved):

PR #297, opened 2023-07-28
[MODUSERS-373]-Added indexes and where clause

SinghAdes:
https://issues.folio.org/browse/MODUSERS-373

### DIFF
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -182,12 +182,6 @@
           "tOps": "ADD",
           "caseSensitive": false,
           "removeAccents": true
-        },
-        {
-          "fieldName": "type",
-          "tOps": "ADD",
-          "caseSensitive": false,
-          "removeAccents": true
         }
       ],
       "fullTextIndex" : [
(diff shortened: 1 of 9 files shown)


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -220,6 +230,10 @@
         {
           "fieldName": "customFields",
           "tOps":"ADD"
+        },
+        {
+          "fieldName" : "type",
+          "tOps" : "ADD"
```
SerhiiNosko, line 236: We don't need fullText and gin index here. 
Just common index, also please make sume testing, create cql query to filter out type='shadow' and make sure that index used.
Firstly you need to check that there is no warning that index missed in mod-users logs.
Secondly you should get generated sql query and invoke EXPLAIN ANALYZE to confirm that index used

SinghAdes: image

SinghAdes: image

SerhiiNosko (approved):

tatsiana-tarhonskaya (approved):

Abdulkhakimov (approved):

gurleenkaurbp (commented):
Looks good

PR #303, opened 2023-08-31
[MODUSERS-384] - Filter out patron users for Consortia pipeline

Dmitriy-Butramyou:
[MODUSERS-384](https://issues.folio.org/browse/MODUSERS-384) - Filter out patron users for Consortia pipeline


src/main/java/org/folio/event/service/UserOutboxService.java
```diff
@@ -219,4 +220,8 @@ private boolean isPersonalNameChanged(Personal oldPersonal, Personal newPersonal
   private boolean isPersonalNotNull(Personal oldPersonal, Personal newPersonal) {
     return ObjectUtils.allNotNull(oldPersonal, newPersonal);
   }
+
+  private boolean isNotPatronUserType(User user) {
+    return ObjectUtils.notEqual(PATRON, user.getType());
```
SerhiiNosko, line 225: I propose to introduce enum UserType with values: patron, staff, shadow

src/test/java/org/folio/rest/impl/UsersAPIIT.java
```diff
@@ -510,6 +510,46 @@ void canDeleteAUser() {
...
+        .build())
+      .tags(TagList.builder().tagList(List.of("foo", "bar")).build())
+      .build();
+
+    final var user = usersClient.createUser(userToCreate);
+
+    usersClient.deleteUser(user.getId());
+
+    List<UserEvent> userCreatedEvents = getUserEventsAndFilterByUserId(USER_CREATED, userId);
+    List<UserEvent> userDeletedEvents = getUserEventsAndFilterByUserId(USER_DELETED, userId);
```
SerhiiNosko, line 543: I think need to add more details maybe in test name and test comments to explain what this test checking.
Like that it checks scenario when we are in consortium mode, and for create events we are not sending events, also need to add some testing for update as well, you can include user update in this test or create new one

src/main/java/org/folio/rest/impl/UsersAPI.java
```diff
@@ -303,6 +304,15 @@ private void saveUser(User entity, Map<String, String> okapiHeaders, PostgresCli
                 "This barcode has already been taken"))));
           return;
         }
+        if (isUserTypeNotPopulatedError(reply)) {
+          logger.info("The user type was not populated for the user with id {}", entity.getId());
```
SerhiiNosko, line 308: We already have similar error log in method isUserTypePopulated()
So this info log can be deleted

src/main/java/org/folio/rest/impl/UsersAPI.java
```diff
@@ -324,6 +334,14 @@ private boolean isDuplicateBarcodeError(String errorMessage) {
     return errorMessage.contains(BARCODE_ALREADY_EXISTS);
   }
 
+  private boolean isUserTypeNotPopulatedError(String errorMessage) {
+    return errorMessage.contains(USER_TYPE_NOT_POPULATED);
+  }
+
+  private boolean isUserTypeNotPopulatedError(AsyncResult<Response> reply) {
+    return isDesiredError(reply, ".*user type was not populated.*");
+  }
```
SerhiiNosko, line 343: lets move these methods after 2 method related to barcodes(and not between them)

src/test/java/org/folio/rest/impl/UsersAPIConsortiaTest.java
```diff
@@ -0,0 +1,366 @@
...
+    addressTypesClient.deleteAllAddressTypes();
+    userTenantClient.deleteAllUserTenants();
+  }
+
+  @Test
+  void doNotSendUserCreateUpdateKafkaEvents() {
+    //That test checks scenario when we are in consortium mode, and for create/update events we are not sending events for patron users
+    commitAllMessagesInTopic(TENANT_NAME, USER_CREATED.getTopicName());
+    commitAllMessagesInTopic(TENANT_NAME, USER_UPDATED.getTopicName());
+    commitAllMessagesInTopic(TENANT_NAME, USER_DELETED.getTopicName());
```
SerhiiNosko, line 65: these 3 lines can be moved to BeforeEach

src/test/java/org/folio/rest/impl/UsersAPIConsortiaTest.java
```diff
@@ -0,0 +1,366 @@
...
+    UserTenant userTenant = new UserTenant()
+      .withId(UUID.randomUUID().toString())
+      .withUserId(UUID.randomUUID().toString())
+      .withUsername("user_test").withTenantId("tenant_test").withCentralTenantId("diku");
+
+    userTenantClient.attemptToSaveUserTenant(userTenant);
+
+    String userId = UUID.randomUUID().toString();
+    final User userToCreate = User.builder()
```
SerhiiNosko, line 74: I think you can create 2 methods to create and update user that can be reused

src/main/java/org/folio/event/service/UserTenantService.java
```diff
@@ -129,4 +144,13 @@ private Future<Void> isUsernameUniqueAcrossTenants(String username, String conso
         return Future.succeededFuture();
       }));
   }
+
+  private Future<Void> isUserTypePopulated(User user) {
+    if (Objects.isNull(user.getType())) {
```
SerhiiNosko, line 149: Thinking lets make check more strict, so fo consortia mode lets check that type equals to some of this values: shadow, staff, patron. So we also update error message with possible values.
Also the same validation we have on UI

SerhiiNosko (approved):

Abdulkhakimov (approved):

PR #305, opened 2023-09-05
[MODUSERS-384] - Added system user type

Dmitriy-Butramyou:
[MODUSERS-384](https://issues.folio.org/browse/MODUSERS-384)
Added system user type

### DIFF
--- a/ramls/userTenant.json
+++ b/ramls/userTenant.json
@@ -44,8 +44,8 @@
       "type": "string"
     },
     "consortiumId": {
-      "description": "The consortiumId",
-      "type": "string"
+      "description": "UUID of the consortiumId",
+      "$ref": "raml-util/schemas/uuid.schema"
     }
   },
   "additionalProperties": false,
--- a/ramls/userTenant.json
+++ b/ramls/userTenant.json
@@ -42,6 +42,10 @@
     "externalSystemId": {
       "description": "The externalSystemId of the user's",
       "type": "string"
+    },
+    "consortiumId": {
+      "description": "The consortiumId",
+      "type": "string"
     }
   },
   "additionalProperties": false,
--- a/src/main/java/org/folio/event/service/UserOutboxService.java
+++ b/src/main/java/org/folio/event/service/UserOutboxService.java
@@ -26,6 +26,7 @@ import java.util.Objects;
 import java.util.UUID;
 
 import static org.folio.domain.UserType.STAFF;
+import static org.folio.domain.UserType.SYSTEM;
 
 public class UserOutboxService {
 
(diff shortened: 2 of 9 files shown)


SerhiiNosko (approved):

src/main/java/org/folio/event/service/UserOutboxService.java
```diff
@@ -233,7 +234,8 @@ private boolean isPersonalNotNull(Personal oldPersonal, Personal newPersonal) {
     return ObjectUtils.allNotNull(oldPersonal, newPersonal);
   }
 
-  private boolean isStaffUser(User user) {
-    return Objects.equals(STAFF.getTypeName(), user.getType());
+  private boolean isStaffOrSystemUser(User user) {
```
okolawole-ebsco, line 237: what if user is null?

Dmitriy-Butramyou: When we work in consortia the 'type' field  is required and that validate before that validation.

okolawole-ebsco: I am referring to the `User` object itself not the type property within.

ramls/userTenant.json
```diff
@@ -42,6 +42,10 @@
     "externalSystemId": {
       "description": "The externalSystemId of the user's",
       "type": "string"
+    },
+    "consortiumId": {
+      "description": "The consortiumId",
+      "type": "string"
```
SerhiiNosko, line 48: propose to use uuid type, see example for userId field

src/main/resources/templates/db_scripts/update_user_tenants_table.sql
```diff
@@ -4,7 +4,8 @@ ALTER TABLE user_tenant
   ADD COLUMN IF NOT EXISTS mobile_phone_number text,
   ADD COLUMN IF NOT EXISTS phone_number text,
   ADD COLUMN IF NOT EXISTS barcode text,
-  ADD COLUMN IF NOT EXISTS external_system_id text;
+  ADD COLUMN IF NOT EXISTS external_system_id text,
+  ADD COLUMN IF NOT EXISTS consortium_id text;
```
SerhiiNosko, line 8: uuid type should be here instead of text

src/test/java/org/folio/rest/impl/UsersAPIConsortiaTest.java
```diff
@@ -106,7 +106,7 @@ void doNotSendUserCreateUpdateKafkaEventsForShadowUser() {
   }
 
   @Test
-  void canDeleteAUserForConsortia() {
+  void canDeleteStaffAUserForConsortia() {
```
SerhiiNosko, line 109: please add asserts to UserTenantIT that field 'consortiumId' is populated

src/main/java/org/folio/event/service/UserOutboxService.java
Dmitriy-Butramyou: We use that validation after create/update user, and I think user can't be null here.

src/main/resources/templates/db_scripts/update_user_tenants_table.sql
Dmitriy-Butramyou: Changed to uuid

src/test/java/org/folio/rest/impl/UsersAPIConsortiaTest.java
Dmitriy-Butramyou: Added assert to UserTenantIT

Abdulkhakimov (approved):

SerhiiNosko:
Karate tests passed:
image

PR #310, opened 2023-09-28
MODUSERS-389. Add check to produce user updated event if user type changed

SerhiiNosko:
https://issues.folio.org/browse/MODUSERS-389

### DIFF
--- a/src/test/java/org/folio/rest/impl/UsersAPIConsortiaTest.java
+++ b/src/test/java/org/folio/rest/impl/UsersAPIConsortiaTest.java
@@ -9,7 +9,6 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 import io.vertx.core.json.Json;
-import io.vertx.core.json.JsonObject;
 import io.vertx.junit5.Timeout;
 import io.vertx.junit5.VertxExtension;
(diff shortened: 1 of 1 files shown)


Abdulkhakimov (approved):

src/main/java/org/folio/event/service/UserOutboxService.java
```diff
@@ -238,4 +242,13 @@ private boolean isStaffOrSystemUser(User user) {
     return Objects.equals(STAFF.getTypeName(), user.getType()) ||
       Objects.equals(SYSTEM.getTypeName(), user.getType());
   }
+
+  private boolean isStaffOrSystemUserUpdated(User user, boolean isConsortiaFieldsUpdated, boolean isPersonalDataChanged) {
+    return isStaffOrSystemUser(user)  && (isConsortiaFieldsUpdated || isPersonalDataChanged);
```
Dmitriy-Butramyou, line 247: Here 2 spaces

Dmitriy-Butramyou (approved):

azizbekxm (approved):

src/main/java/org/folio/event/service/UserOutboxService.java
SerhiiNosko: fixed

PR #320, opened 2023-12-06
MODUSERS-408. Additional fix for case insensitive username check

SerhiiNosko:
## Purpose
https://issues.folio.org/browse/MODUSERS-408

### DIFF
--- a/src/main/java/org/folio/repository/UserTenantRepository.java
+++ b/src/main/java/org/folio/repository/UserTenantRepository.java
@@ -28,7 +28,7 @@ public class UserTenantRepository {
   public static final String BARCODE = "barcode";
   public static final String EXTERNAL_SYSTEM_ID = "external_system_id";
   public static final String CONSORTIUM_ID = "consortium_id";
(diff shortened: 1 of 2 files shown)


imerabishvili (approved):

Dmitriy-Butramyou (approved):

src/main/java/org/folio/repository/UserTenantRepository.java
```diff
@@ -28,7 +28,7 @@ public class UserTenantRepository {
   public static final String BARCODE = "barcode";
   public static final String EXTERNAL_SYSTEM_ID = "external_system_id";
   public static final String CONSORTIUM_ID = "consortium_id";
-  public static final String LOWERCASE_WRAPPED_USERNAME = String.format("LOWER(%s)", USERNAME_FIELD);
+  public static final String LOWERCASE_WRAPPED_USERNAME = String.format("LOWER(f_unaccent(%s))", USERNAME_FIELD);
```
SerhiiNosko, line 31: Modified this query to use indexes, for testing I created 1k records on rancher and run this query without f_unaccent
image
And the same query but with adding f_unaccent
image

azizbekxm (approved):

PR #327, opened 2023-12-21
MODUSERS-408. Forbid request to ECS tenant if it soft deleted

SerhiiNosko:
## Purpose
https://issues.folio.org/browse/MODUSERS-411

### DIFF
--- a/src/main/java/org/folio/event/service/UserTenantService.java
+++ b/src/main/java/org/folio/event/service/UserTenantService.java
@@ -35,8 +35,7 @@ public class UserTenantService {
   private static final Logger logger = LogManager.getLogger(UserTenantService.class);
   public static final String INVALID_USER_TYPE_POPULATED = "User's 'type' field should be populated with one of the allowed values: 'patron', 'staff', 'shadow'";
   public static final String USERNAME_IS_NOT_POPULATED = "In consortium mode, the staff user must have a username";
-  public static final String MEMBER_USER_TENANT_TABLE_IS_EMPTY = "User-tenant table is empty, nothing to delete";
-  public static final String MEMBER_USER_TENANT_CONTAINS_ADDITIONAL_RECORDS = "User-tenant table in member tenant should not contain more than 1 record";
+  public static final String MEMBER_USER_TENANT_SHOULD_CONTAIN_SINGLE_RECORD = "User-tenant table in member ECS tenant should contain 1 record";
 
   private final UserTenantRepository tenantRepository;
   private final BiFunction<Vertx, String, PostgresClient> pgClientFactory;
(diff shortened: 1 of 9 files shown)


src/main/java/org/folio/event/service/UserTenantService.java
```diff
@@ -75,11 +76,41 @@ public Future<String> getConsortiaCentralTenantId(Conn conn, Map<String, String>
...
+   * @param tenantId the tenant id
+   * @param vertx the vertx instance
+   * @return future with true if record was deleted or false otherwise
+   * @throws IllegalStateException if deleting is not possible
+   */
+  public Future<Boolean> deleteMemberUserTenant(String tenantId, Vertx vertx) {
+    PostgresClient pgClient = pgClientFactory.apply(vertx, tenantId);
+    return pgClient.withConn(conn -> fetchFirstUserTenant(conn, tenantId)
+      .compose(res -> {
+        if (res.getTotalRecords() > 1) {
```
imerabishvili, line 98: If we don't have any additional details them maybe:
```
if total_records != 0 then
  throw "should contains exactly one record";
}
```

src/main/java/org/folio/rest/impl/UserTenantsAPI.java
```diff
@@ -76,6 +76,20 @@ public void postUserTenants(String lang, UserTenant userTenant, Map<String, Stri
       });
   }
 
+  @Override
+  public void deleteUserTenants(Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
+    String okapiTenantId = TenantTool.tenantId(okapiHeaders);
+    userTenantService.deleteMemberUserTenant(okapiTenantId, vertxContext.owner())
```
imerabishvili, line 82: Current contract for `deleteMemberUserTenant` is that it successfully deletes exactly one record or throws an exception. Let's make it return Future<Void> then to avoid ambiguity. We can replace all 'boolean's with `void` in the chain.

src/main/java/org/folio/event/service/UserTenantService.java
SerhiiNosko: Updated, it makes sence for me as well to simplify, thanks

src/main/java/org/folio/rest/impl/UserTenantsAPI.java
SerhiiNosko: basically internally deleteMemberUserTenant invokes DAO method that also can return true or false, so we can not just simply return Future<Void> in this case, I also wrote java doc to explain contract more

imerabishvili: Can `deleteMemberUserTenant` return `false`?

imerabishvili: I think it can not. Do we handle a case if `deleteById` returns false?

SerhiiNosko: Added check for dao method as well

src/main/java/org/folio/event/service/UserTenantService.java
```diff
@@ -75,11 +75,44 @@ public Future<String> getConsortiaCentralTenantId(Conn conn, Map<String, String>
...
+   * (central ECS tenant contains multiple user tenant associations necessary for login, saml-login, forgot password/username functionality).
+   * This method deletes  record from user-tenant table in Member ECS tenant and after this each request to that tenant will be
+   * forbidden.
+   *
+   * @param tenantId the tenant id
+   * @param vertx the vertx instance
+   * @return future with true if record was deleted or false otherwise
+   * @throws IllegalStateException if deleting is not possible
+   */
+  public Future<Boolean> deleteMemberUserTenant(String tenantId, Vertx vertx) {
```
imerabishvili, line 93: `Boolean` is not used (we always return true). Can be `Future<Void>`.

imerabishvili (approved):

azizbekxm (approved):

PR #404, opened 2025-03-05
MODUSERS-507. Staff User created in member tenant becomes as Shadow after updating email, firstname

SerhiiNosko:
## Purpose
https://folio-org.atlassian.net/browse/MODUSERS-507

- Shadow users are copies of real staff users, skip sending update events for them

- Remove user info from logs

- Close kafka producers to avoid memory leaks

### DIFF
--- a/src/main/java/org/folio/service/event/DomainEventPublisher.java
+++ b/src/main/java/org/folio/service/event/DomainEventPublisher.java
@@ -48,20 +48,23 @@ public class DomainEventPublisher<K, T> {
         .topic(kafkaTopic)
         .propagateOkapiHeaders(okapiHeaders)
         .build();
-    log.info("publish:: kafkaRecord = [{}]", producerRecord);
 
     KafkaProducer<K, String> producer = getOrCreateProducer();
     log.info("publish:: Producer created, sending the record...");
 
     return producer.send(producerRecord)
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/service/event/DomainEventPublisher.java
```diff
@@ -48,20 +48,23 @@ public Future<Void> publish(K key, DomainEvent<T> event, Map<String, String> oka
...
-      .onSuccess(r -> log.info("publish:: Succeeded sending domain event with key [{}], " +
-        "kafka record [{}]", key, producerRecord))
-      .<Void>mapEmpty()
-      .onFailure(cause -> {
-        log.error("publish:: Unable to send domain event with key [{}], kafka record [{}]",
-          key, producerRecord, cause);
-        failureHandler.handle(cause, producerRecord);
-      });
+      .onComplete(r -> {
+        producer.end(ear -> producer.close());
```
SerhiiNosko, line 57: Missing closing producer will cause memory leak like this: https://folio-org.atlassian.net/browse/MODINVSTOR-794
All other places make closing of kafka producer: 
https://github.com/search?q=org%3Afolio-org+producer.close%28%29&type=code

BKadirkhodjaev (approved):

azizbekxm (approved):

gurleenkaurbp (approved):

